PR #15, opened 2020-10-02
feat: add fastly provider

btkostner:
This adds a [fastly image optimization](https://docs.fastly.com/en/guides/image-optimization-api) provider.

- I tried to get the size operator matched up as much as I could, but they only have three options. 
- I added a fastly url, but because it's a demo url, the only available image to test on is [`/image.jpg`](https://www.fastly.io/image.jpg)

### DIFF
--- a/docs/content/en/providers.md
+++ b/docs/content/en/providers.md
@@ -43,8 +43,6 @@ export default {
 }
 ```
 
-
-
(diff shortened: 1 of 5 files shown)


farnabaz (approved):
Great, Thanks 👍

PR #17, opened 2020-10-03
docs: give <img> empty alt="" attribute

dennishzg:
Improved accessibility. Added empty alt="" attributes to remove the decorative images from the accessibility tree.

### DIFF
--- a/docs/content/en/index.md
+++ b/docs/content/en/index.md
@@ -9,8 +9,8 @@ features:
   - CDN Support
 ---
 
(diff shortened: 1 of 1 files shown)


farnabaz:
Thank you for PR 
Nuxt Image automatically generates `alt` attribute from the file name if it is empty. Setting an empty `alt` attribute is redundant therefore, I'm closing this issue.   

The docs website is outdated. That's why there is no alt attribute in images.

PR #31, opened 2020-10-11
test: update ci and tests

farnabaz:
It seems that playwright and githucCI, The cached files of playwright do not restore on later builds. The solution is to cache yarn cache directory instead of node_modules.

check out https://github.com/nuxt/image/pull/22/checks?check_run_id=1216330529


pi0:
(PR description is مستحب :D)

farnabaz:
@pi0 (on my eyes)
Check out the description

pi0:
Alternatively, we can have a parallel action for browser/e2e tests without cache (or caching global yarn cache dir): https://github.com/nuxt/test-utils/blob/master/.github/workflows/e2e.yml

Issue with playwright is that it depends on a postinstall script that when we restore cache, is not triggered to download native deps..

PR #37, opened 2020-10-14
feat: auto detect image ratio

### LINKED ISSUE #30: Concerns regarding fixed layout sizing
From what I read in the doc and what I experiment, the `width` and `height` attributes are mandatory. This is a good thing for the CLS reasons explained in the doc but I think they are currently misused. The main reason we have to put `width` and `height` on `img` elements is to inform the browsers about the aspect ratio. 

Here is a quote of Addy Osmani on the subject (full article [here](https://web.dev/optimize-cls/)).

> Modern browsers now set the default aspect ratio of images based on an image's width and height attributes so it's valuable to set them to prevent layout shifts. Thanks to the CSS Working Group, developers just need to set width and height as normal and the UA stylesheets of all browsers add a default aspect ratio based on the element's existing width and height attributes.

In the current implementation, the `width` is used to define the width of the container which in responsive development is a **blocking behavior**. I think developers should be responsible for sizing their items while `nuxt-image` should adopt a responsive approach with relative sizings.

Eg: [https://codepen.io/twicpics/pen/627fd52cd4dc523ff72c2fadef9b78a0?editors=1100](https://codepen.io/twicpics/pen/627fd52cd4dc523ff72c2fadef9b78a0?editors=1100)

Then let browsers choose the correct image by playing with `srcset`.

Eg: [https://codepen.io/twicpics/pen/b87821d341540afe22cd079059536f84?editors=1100](https://codepen.io/twicpics/pen/b87821d341540afe22cd079059536f84?editors=1100)

Regarding the placeholder behavior, same idea: the `padding-top` trick could be dynamically computed by using CSS `calc()`. 

```css
.placeholder {
  padding-top: calc(height / width * 100%);
}
```
Eg: [https://codepen.io/twicpics/pen/d6317ba5001ecdd3f2920d7dfb2e782a?editors=1100](https://codepen.io/twicpics/pen/d6317ba5001ecdd3f2920d7dfb2e782a?editors=1100)

I didn't dig into the "fit" attribute but it shouldn't be a problem.

Does it make sens?

farnabaz:
Using the feature introduced in https://github.com/nuxt-contrib/ipx/pull/8 we can detect image original width and height and calculate image ratio to prevent CLS and remove explicit image width.

Fix #30


docs/providers/random/runtime.js
```diff
@@ -1,5 +1,5 @@
 export default {
-  generateURL (src, modifiers, options) {
+  getItem (src, modifiers, options) {
```
Atinux, line 2:
```suggestion
  getImage (src, modifiers, options) {
```

src/index.ts
```diff
@@ -151,5 +135,34 @@ function handleStaticGeneration (nuxt: any) {
...
+  /* istanbul ignore if */
+  if (defaultHost === '0.0.0.0') {
+    defaultHost = 'localhost'
+  }
+
+  // Default prefix
+  const prefix = '/'
+
+  return defu(providerOptions, {
+    baseURL: `http://${defaultHost}:${defaultPort}${prefix}`,
```
pi0, line 162: - baseURL in browser may differ from `HOST` when deployed to production
- For server-side, we may directly listen to a random port 

In conclusion, it would be best if at least support two baseURL options and improve later

src/providers/fastly/runtime.ts
```diff
@@ -16,10 +16,21 @@ const operationsGenerator = createOperationsGenerator({
...
 export default <RuntimeProvider> {
-  generateURL (src: string, modifiers: ImageModifiers, options: any) {
+  getImage (src: string, modifiers: ImageModifiers, options: any) {
     const operations = operationsGenerator(modifiers)
+    const url = cleanDoubleSlashes(options.baseURL + src + '?' + operations)
     return {
-      url: cleanDoubleSlashes(options.baseURL + src + '?' + operations)
+      url,
+      getInfo: async () => {
+        const infoString = await fetch(url).then(res => res.headers.get('fastly-io-info') || '')
```
pi0, line 25: A tip for later: We may introduce a fetch utility (imported from `~image/fetch`) that can also do an in memory LRU cache

src/providers/local/index.ts
```diff
@@ -12,10 +14,18 @@ function createMiddleware (options) {
...
-      dir: options.dir
-    },
+    inputs: [
+      {
+        name: 'local',
+        adapter: 'fs',
+        dir: options.dir
+      },
+      {
+        name: 'remote',
```
pi0, line 24: We should somehow guard this to only enable during `generate` to avoid service abuse

src/runtime/image.ts
```diff
@@ -88,6 +89,26 @@ export function createImage (context, { providers, defaultProvider, presets }: C
...
   })
 
+  image.lqip = async (source: string, options: any = {}) => {
+    const { src, provider: sourceProvider } = processSource(source)
+    const provider = providers[sourceProvider || options.provider || defaultProvider]
+    if (!provider) {
+      throw new Error('Unsupported provider ' + provider)
+    }
+
+    const generated = provider.provider.getImage(src, {
```
pi0, line 99:
```suggestion
    const image = provider.provider.getImage(src, {
```

src/runtime/image.ts
```diff
@@ -88,6 +89,26 @@ export function createImage (context, { providers, defaultProvider, presets }: C
     }
   })
 
+  image.lqip = async (source: string, options: any = {}) => {
```
pi0, line 92: What if we use simpler `placeholder` keyworkd?

src/runtime/image.ts
```diff
@@ -88,6 +89,26 @@ export function createImage (context, { providers, defaultProvider, presets }: C
...
+    const provider = providers[sourceProvider || options.provider || defaultProvider]
+    if (!provider) {
+      throw new Error('Unsupported provider ' + provider)
+    }
+
+    const generated = provider.provider.getImage(src, {
+      width: 30
+    }, provider.defaults)
+
+    if (typeof generated.getInfo !== 'function') {
```
pi0, line 103: Maybe we can support `.getPlaceholder` to let providers have their own logic as well?

src/runtime/nuxt-image-mixins.ts
```diff
@@ -68,13 +68,23 @@ export default {
...
-      this.blurry = await this.$img.lqip(this.src)
+    if (this.legacy) {
+      return
     }
+    const image = await this.$img.lqip(this.src)
+    if (!image) {
+      return
+    }
+
+    this.blurry = image.url
```
pi0, line 85: Same here. We are using different terms for lqip while it can be simply called placeholder everywhere

src/runtime/nuxt-image-mixins.ts
```diff
@@ -68,13 +68,23 @@ export default {
...
+    this.imageW = image.width
```
pi0, line 86: `this.placeholderWidth` ?

src/runtime/nuxt-image-mixins.ts
```diff
@@ -85,6 +95,18 @@ export default {
...
   },
   computed: {
+    imageWidth () {
+      return this.width || this.imageW
+    },
+    imageHeight () {
+      return this.height || this.imageH
+    },
+    imageRatio () {
+      if (this.height && this.width) {
```
pi0, line 105: Can we use new `imageWidth` computed to merge to conditions?

src/runtime/provider-utils.ts
```diff
@@ -25,7 +25,7 @@ export function createOperationsGenerator ({ formatter, keyMap, joinWith = '/',
   })
 
   return (modifiers: { [key: string]: any }) => {
-    const operations = Object.entries<string>(modifiers)
+    const operations = Object.entries(modifiers)
```
pi0, line 28: (Can we make type to be `: string` ? :D)

types/runtime.d.ts
```diff
@@ -48,7 +48,22 @@ export interface RuntimeProvider {
...
+  getInfo?: () => Promise<RuntimeImageInfo>
+}
+
+export interface RuntimeImageInfo {
+  // width of image in pixels
+  width: number,
+  // height of image in pixels
+  height: number,
+  // size of image in bytes
+  size: number,
```
pi0, line 66: What if call it `bytes` to not confuse with px size and also making self explain that unit is bytes?

pi0 (commented):
Nice additions! Sorry for my verbose comments :see_no_evil:

src/runtime/nuxt-image-mixins.ts
farnabaz: It isn't safe to use `imageWidth` to calculate ratio, it leads to wrong ratio if user provides `width` prop and does not provide `height` prop

src/runtime/nuxt-image-mixins.ts
farnabaz: It is the original width of the image, how about `this.originalWidth`

src/runtime/nuxt-image-mixins.ts
pi0: But then Is it useful if user only sets one of `width` or `height`? (if not we may throw an error in created about wrong usage)

farnabaz: I do refactor `imageRatio` property and manage to remove `imageWidth` & `imageHeight` computed because they have no real use in the component.
`width` and `height` are only used for calculating ratio and they are optional, missing one of them is like missing both of them. 
I think missing one of these props is not a critical error

src/providers/local/index.ts
farnabaz: We can deny all remote urls by default, and If the user wants to use a remote url she should add domains to module whitelist.
What do you think

src/index.ts
farnabaz: I have created `internalBaseURL` option for server-side usage. [`4bd3543` (#37)](https://github.com/nuxt/image/pull/37/commits/4bd3543aa2bb5d9bc0addaf8e73be39c2d8cb181)

src/runtime/nuxt-image-mixins.ts
pi0: > width and height are only used for calculating ratio and they are optional, missing one of them is like missing both of them

I think we should at least warn users that prop is ignored instead of silently ignoring wrong configuration

pi0 (approved):

PR #43, opened 2020-10-19
feat(runtime): catch image exceptions and prevent page crash

farnabaz:
Doing wrong with `nuxt-image` should not lead page crash. When the user uses an invalid provider or preset Image module throw an exception, We need to catch these exceptions to prevent page crash.  
Instead we can show the exception message inside `nuxt-image`.

Screen Shot 2020-10-19 at 12 58 57 PM

### DIFF
--- a/src/runtime/nuxt-image-mixins.ts
+++ b/src/runtime/nuxt-image-mixins.ts
@@ -149,7 +149,7 @@ export default {
       try {
         return this.$img.lqip(this.src)
       } catch (e) {
-        this.error = e.message
+        this.onError(e)
         return false
       }
     },
@@ -164,7 +164,8 @@ export default {
         })
         return encodeURI(image)
       } catch (e) {
-        this.error = e.message
+        this.onError(e)
+        return ''
       }
     },
     loadOriginalImage () {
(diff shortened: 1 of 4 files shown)


src/runtime/nuxt-image-mixins.ts
```diff
@@ -144,15 +145,27 @@ export default {
     }
   },
   methods: {
+    getPlaceholder () {
+      try {
+        return this.$img.lqip(this.src)
+      } catch (e) {
+        this.error = e.message
```
pi0, line 152: - We may unify to `onError` method
- We are hiding stack-trace, would be nice to also `console.error` original error

src/runtime/nuxt-image-mixins.ts
```diff
@@ -65,6 +65,7 @@ export default {
   },
   data () {
     return {
+      error: '',
```
pi0, line 68: Is the type object or string? :D

pi0 (approved):

src/runtime/nuxt-image-mixins.ts
farnabaz: I store error message not whole error object `this.error = e.message`.
Do we need error object? 🤔

PR #48, opened 2020-10-22
feat: configurable placeholder type

farnabaz:
This PR works with https://github.com/nuxt-contrib/ipx/pull/10 to support variant types for placeholder image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -163,7 +163,11 @@ function prepareLocalProvider ({ nuxt, options }, providerOptions) {
   providerOptions = defu(providerOptions, {
     baseURL: `http://${defaultHost}:${defaultPort}${prefix}`,
     internalBaseURL: `http://${defaultHost}:${defaultPort}${prefix}`,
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -163,7 +163,11 @@ function prepareLocalProvider ({ nuxt, options }, providerOptions) {
   providerOptions = defu(providerOptions, {
     baseURL: `http://${defaultHost}:${defaultPort}${prefix}`,
     internalBaseURL: `http://${defaultHost}:${defaultPort}${prefix}`,
-    dir: path.resolve(nuxt.options.srcDir, nuxt.options.dir.static)
+    dir: path.resolve(nuxt.options.srcDir, nuxt.options.dir.static),
+    placeholder: {
+      type: 'sqip',
```
farnabaz, line 168: I think it would better if we use `image` as the default placeholder type and if make sqip a peer dependency.
What do you think?

PR #55, opened 2020-10-28
feat: universal meta resolving

pi0:
Previously we was depending on providers to resolve image meta (getting aspect ratio and also base64) this was adding complexity and sometimes extera requests. With this PR, using [nuxt-contrib/image-meta](https://github.com/nuxt-contrib/image-meta) (for Node.js) and `Image` + `Canvas` (for Browser), we can universally have ratio. (Future release of `image-meta` may eventually support Node.js if we can remove Buffer dependency)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "defu": "^3.1.0",
     "hasha": "^5.2.2",
-    "image-size": "^0.9.3",
+    "image-meta": "^0.0.1",
     "ipx": "0.4.0-rc.1",
     "node-fetch": "^2.6.1",
     "upath": "^2.0.0"
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "hasha": "^5.2.2",
     "image-meta": "^0.0.1",
     "ipx": "0.4.0-rc.1",
+    "make-fetch-happen": "^8.0.10",
     "node-fetch": "^2.6.1",
     "upath": "^2.0.0"
   },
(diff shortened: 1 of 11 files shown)


src/index.ts
```diff
@@ -102,7 +102,7 @@ function loadProvider (key: string, provider: any) {
 function handleStaticGeneration (nuxt: any) {
   const staticImages = {} // url ~> hash
 
-  nuxt.hook('vue-renderer:ssr:prepareContext', (renderContext) => {
+  nuxt.hook('vue-renderer:ssr:prepareContext', async (renderContext) => {
```
pi0, line 105: Unwanted change?

farnabaz: yes

farnabaz (approved):

src/index.ts
```diff
@@ -117,6 +123,7 @@ function handleStaticGeneration (nuxt: any) {
 
   nuxt.hook('generate:done', async () => {
     const { port } = nuxt.server.listeners[0]
+    console.log("345678", nuxt.server.listeners[0])
```
pi0: ??

package.json
```diff
@@ -27,6 +27,7 @@
     "hasha": "^5.2.2",
     "image-meta": "^0.0.1",
     "ipx": "0.4.0-rc.1",
+    "make-fetch-happen": "^8.0.10",
```
pi0, line 30: Are we using this package?

PR #64, opened 2020-11-03
refactor image utility

farnabaz:
With this PR:

- Image utility has sizes operation
- Module listen to a random port for internal usage
- Image utility detects client-side rendering and fallback to the original image on static images
- Rename data property to `nuxtImageMap` in order to reduce conflict possibility with users data
- Use default sizes if the user does not provide sets

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -1,5 +1,6 @@
-import type { CreateImageOptions, ImageModifiers, ImagePreset } from 'types'
+import type { CreateImageOptions, ImageModifiers, ImagePreset, ImageSize } from 'types'
 import { getMeta } from './meta'
+import { cleanDoubleSlashes } from './utils'
 
 function processSource (src: string) {
   if (!src.includes(':') || src.match('^https?://')) {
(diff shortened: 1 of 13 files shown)


src/runtime/image.ts
```diff
@@ -56,7 +57,7 @@ export function createImage (context, { providers, defaultProvider, presets, int
     return presetMap[name]
   }
 
-  function image (source: string, modifiers: ImageModifiers, options: any = {}) {
+  function parseImage (source: string, modifiers: ImageModifiers, options: any = {}) {
```
pi0, line 60: I would suggest naming `createImage` since we do more than parsing and basically creating an image object

pi0 (approved):

src/runtime/nuxt-image-mixins.ts
```diff
@@ -78,6 +78,11 @@ export default {
   },
   async fetch () {
     await this.fetchMeta()
+
+    if (this.$nuxt.context.ssrContext && this.$nuxt.context.ssrContext.isGenerating) {
+      // eslint-disable-next-line no-unused-expressions
+      this.sizes
```
pi0, line 84: ??

farnabaz: @pi0  Oh, I forgot to put a comment above this.
In the `nuxt generate` we need to generate image srcsets and download them into `dist` directory, but because of lazy loading, the sizes does not calculate on the server side. With this we make sure that sizes are generates and downloads on static build.

PR #66, opened 2020-11-04
feat: load all images on print

### LINKED ISSUE #56: Load Images on Print
Ref implementation: https://github.com/mfranzke/loading-attribute-polyfill/blob/b1ebfbe0531597f3aea559ee03c06a45a1dc8f4f/loading-attribute-polyfill.js#L128

image

farnabaz:
Use `window.matchMedia` to detect print media query and load all images on the page.

close #56


src/runtime/image.ts
```diff
@@ -137,36 +137,61 @@ export function createImage (context, { providers, defaultProvider, presets, int
...
     rootMargin: '50px',
     ...options
-  }
-  const observer = (typeof IntersectionObserver !== 'undefined' ? new IntersectionObserver(callback, intersectOptions) : {}) as IntersectionObserver
+  }) : {}) as IntersectionObserver
+
+  return observer
+}
+
+function createObserver (intersectionOptions: object) {
```
pi0, line 162: I think for longer term it would be better to create observer directly in component instead of share runtime logic (`createObserver(image)`)

pi0 (approved):

src/runtime/image.ts
farnabaz: If we move observer into component we are creating multiple query match and IntersectionOsberver. I'm not sure but it may reduce the performance

PR #67, opened 2020-11-05
feat: inline styles

farnabaz:
Remove css file and inline styles, This way we can have more control over animation and styling on runtime. Also image componet does not wait for styles from resource file.


pi0 (approved):
Much cleaner 👍

PR #76, opened 2020-11-11
fix(cloudinary): fit types, auto format & quality

mayashavin:
Add the following:
1. More types for resizing "fit"
2. Add default format and quality
3. Remove extension for Cloudinary images since they are not needed.
4. Fix unit tests.

### DIFF
--- a/src/providers/cloudinary/runtime.ts
+++ b/src/providers/cloudinary/runtime.ts
@@ -42,7 +42,7 @@ export default <RuntimeProvider> {
 
     const srcWithoutExtension = src.replace(/\.[^/.]+$/, '')
     const operations = operationsGenerator(mergeModifiers as ImageModifiers)
-    
+
     return {
       url: cleanDoubleSlashes(options.baseURL + '/' + operations + srcWithoutExtension)
     }
--- a/src/providers/cloudinary/runtime.ts
+++ b/src/providers/cloudinary/runtime.ts
@@ -6,7 +6,8 @@ const operationsGenerator = createOperationsGenerator({
     fit: 'c',
     width: 'w',
     height: 'h',
-    format: 'f'
+    format: 'f',
+    quality: 'q'
   },
   valueMap: {
     fit: {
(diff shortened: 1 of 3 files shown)


src/providers/cloudinary/runtime.ts
```diff
@@ -6,27 +6,45 @@ const operationsGenerator = createOperationsGenerator({
...
 export default <RuntimeProvider> {
   getImage (src: string, modifiers: ImageModifiers, options: any) {
-    const operations = operationsGenerator(modifiers)
+    const mergeModifiers = {
+      ...modifiers,
+      format: modifiers.format || defaultModifiers.format,
+      quality: modifiers.quality || defaultModifiers.quality
+    }
+
+    const srcWithoutExtension = src.replace(/\.[^/.]+$/, '')
```
pi0, line 43: What does happens if two images uploaded on same path and different format?

mayashavin:
What do you mean :)? There are no two images uploaded on the same path
without overriding each other, their unique public id is different and
decided by the file name, not the extension. In other words, if you upload
an image called 'kitten.png', and then upload 'kitten.jpg', the second one
will replace the first one in Cloudinary if there is `overwritten` enabled.
Otherwise, the second one will be saved with different name, or a random
public id, or `kitten_1` maybe, depending on how you configure it.

Let me know if you need anything.

Best regards,
*Maya Shavin*
Senior FrontEnd Developer
📱*:* +972-54-9445592 <+972%2054-9445592>
www.cloudinary.com

src/providers/cloudinary/runtime.ts
mayashavin: There are no two images uploaded on the same path
without overriding each other, their unique public id is different and
decided by the file name, not the extension. In other words, if you upload
an image called 'kitten.png', and then upload 'kitten.jpg', the second one
will replace the first one in Cloudinary if there is `overwritten` enabled.
Otherwise, the second one will be saved with different name, or a random
public id, or `kitten_1` maybe, depending on how you configure it. :)

pi0: Good to know. Then it is safe :)

src/providers/cloudinary/runtime.ts
```diff
@@ -6,27 +6,45 @@ const operationsGenerator = createOperationsGenerator({
...
 
+const defaultModifiers = {
+  format: 'auto',
+  quality: 'auto'
+}
+
 export default <RuntimeProvider> {
   getImage (src: string, modifiers: ImageModifiers, options: any) {
-    const operations = operationsGenerator(modifiers)
+    const mergeModifiers = {
```
pi0, line 37: Can be `modifiers = { ...defaultModofiers, ...modifiers }` 😊

mayashavin: Would love to do it. Actually I tried it before, the problem is in `image.ts`, `format` is passed with `undefined` value 
```js
if (!size.format) {
  size.format = operations.format
}
```

And here : 
```js
const { url } = image(src, {
  ...operations,
  width: size.width,
  format: size.format
})
```

so it always overwrite the default value with `undefined` if there is none passed. Either we fix the code logic here, or we check per provider :). WDYT?

pi0: I think defaults should be `null` then we can use `defu` here /cc @farnabaz

farnabaz: As we are refactoring components, This can be fixed in the refactor process

farnabaz (approved):

farnabaz:
Thank you @mayashavin ❤️

PR #79, opened 2020-11-11
fix: improve error readability

### LINKED ISSUE #77: Error not readable when url is not valid - Attempt to write outside buffer bounds
When the generated url is not valid for any reason (wrong modifier passed, wrong format, etc), it throws the error on the call `data.buffer()` - `Attempt to write outside buffer bounds`. The error is not clear what went wrong.

Expected: A more readable message to inform user about the invalid url.

farnabaz:
close #77


farnabaz:
Thank you @pi0 ❤️

PR #98, opened 2020-11-16
Improve providers

### LINKED ISSUE #81: [dx] top level provider options
```js
export default {
 image: {
    cloudinary: {}
 }
 random: {
   provider: '~~/providers/random',
   options: {

   }
 }
}
```

### LINKED ISSUE #83: [dx] rename ìpx`back to local
or alias..

### LINKED ISSUE #84: support `NUXT_IMAGE_PROVIDER` environment variable
to override options.provider` (we may also auto detect it based on deployment target)

### LINKED ISSUE #97: Non default base path breaks nuxt image
When using a non-default base path (i.e. `/`) then the components stops working.

If this is a configuration issue, would it please be possible to describe in the docs how to modify the setup to work with non-default base paths?

Thanks.

farnabaz:
- rename ipx to local (close #83)
- clean up double slashes outside of provider runtime
- handle router base (close #97)
- support `NUXT_IMAGE_PROVIDER` environment variable (close #84)
- allow top-level providers config (close #81)
official providers config can be defined at top-level of `image` config and custom providers should define uneder `providers` property

```js
export default { 
  image: {
     cloudinary: {
       // config
     },
     providers: {
        yourAwesomeProvider: {
          provider: '~/providers/yourAwesomeProvider',
          options: {
            // config
          }
        }
     }
  }

```


enbo:
> * handle router base (close https://github.com/nuxt/image/issues/97)

Can someone refer me to documentation on how to correctly handle a non-default router base in v0 / Nuxt 2? `@nuxt/image@0.7.1` doesn't seem to prepend the router base set in `nuxt.config.js` at all when generating a static dist.

PR #109, opened 2020-12-02
feat: imagekit provider implementation and tests

Alucard17:
- Added Imagekit Provider according to [docs](https://docs.imagekit.io/features/image-transformations/resize-crop-and-other-transformations).
- Added Test for Imagekit provider.
- Added an Example of imagekit in playround.


pi0:
Thanks @Alucard17

PR #110, opened 2020-12-02
Cloudinary documentation, examples, extra operations and some minor fixes

mayashavin:
## Changes:

1. Fix examples of Cloudinary on playground.
2. Add documentation for Cloudinary providers.
3. Add additional mapping for operations.
4. Capitalize name of providers in docs, and fix minor typos.
5. Probably that's it :)

### DIFF
--- a/docs/content/en/nuxt-img.md
+++ b/docs/content/en/nuxt-img.md
@@ -251,9 +251,15 @@ There are five standard values you can use with this property.
   </code-block>
 </code-group>
 
+<alert type="info">
+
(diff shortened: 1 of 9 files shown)


pi0 (approved):
Thanks 🙌

Some changes might add after dev refractors:
- Moving operation docs into new section (not inside nuxt-img)
- Remove `provider=` attribute from example in favor of global provider option
- Creating an operations table to compare provider support
- Prefix vendor specific operations with something like cf- or x- or v-

PR #123, opened 2020-12-21
fix: avoid direct access to window.$nuxt._pagePayload

StephanGerbeth:
Currently the $nuxt._pagePayload is accessed directly on the client side. However, _pagePayload is overwritten by payload prefetches (resourceHints). This prevents lazy hydrated components from accessing the value.

**Solution:**
When a route is changed, the payload is written away via beforeNuxtRender and can be read by lazy hydrated components at a later time.


pi0:
Hi @StephanGerbeth. Would you please rebase your branch with `dev`?

PR #124, opened 2020-12-21
Accessibility: allow empty alt attributes 

bdekok:
For accessibility it is [often a good practice to have an image with an empty alt attribute](https://www.w3.org/WAI/tutorials/images/decision-tree/), however in the current implementation of Nuxt-image it replaces the empty alt for a generated alt. 

This commit changes the behaviour so that if an alt attribute is defined by the user it will always use that, if no alt attribute is defined it will use the generated alt just like before. I have ran the tests and everything seems to be green, if you'd like I can add a test specific for this.

### DIFF
--- a/src/runtime/nuxt-image-mixins.ts
+++ b/src/runtime/nuxt-image-mixins.ts
@@ -49,7 +49,7 @@ export default {
     // `<img>` attrubutes
     alt: {
       type: String,
-      default: ''
+      default: undefined
     },
     referrerpolicy: {
       type: String,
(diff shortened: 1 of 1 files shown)


bdekok:
I spotted that the same behaviour is also on your dev branch, can also make the change on that one.

pi0:
Thanks for PR actually it totally makes sense. Just sicne there are losts of changes in dev/dev-refactors branches, i'm closing this PR but will apply patch manually on top of refactors. Thanks for PR @bdekok <3

PR #133, opened 2020-12-30
feat: add storyblok provider

f3ltron:
Storyblok provider based on storyblok image service documentation 

https://www.storyblok.com/docs/image-service

Added storyblok provider
Added documentation


src/providers/storyblok/runtime.ts
```diff
@@ -0,0 +1,54 @@
...
+  return ret
+}
+
+export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = 'https://img2.storyblok.com' } = {}) => {
+  const mergeModifiers = overrideUndefined({
+    ...defaultModifiers,
+    ...modifiers
+  })
+
+  const srcWithoutBase = src.replace('https://a.storyblok.com', '')
```
pi0, line 48: Is there a reason to unconditionally trim `a.storyblock.com` from src?

f3ltron: Yes all urls from storyblok come like this and need to be change by the cdn url. https://www.storyblok.com/docs/image-service

pi0: Thanks. Would you please add a comment above linking to docs for replacing?

pi0: - we probably also need to ensure not adding duplicate prefix is src is img2

- Also other providers use resource address instead of full URL. This feature seems little bit inconsistency

f3ltron: - Not sure to understand the first point
- Yes but it really depends on the provider :s or I have to transform it with methods before passing it to the component

tillmon:
The image dimension should be included for aspect ratio calculation. See https://www.storyblok.com/faq/image-dimensions-assets-js

```
const url = src

const dimensions = {
    width: url.split('/')[5].split('x')[0],
    height: url.split('/')[5].split('x')[1]
}

return {
    url
    getInfo: () => {
        return {
            width: dimensions.width,
            height: dimensions.height
        }
    }
}
```

tillmon:
Also the docs should be updated.

f3ltron:
Thank's for your feedback @tillmon. I am not sure to understand why I have to use it as the dimension of the image will be calculated by stoyblok automatically? Can you elaborate on why I should add it, please?

tillmon:
I thought this is needed to calculate the space the image is occupying before it is loaded, but I am not sure.

f3ltron:
I am pretty sure it's not useful. 
User can define the default size or let storyblok calculate it. I don't think there is use case for it. But if people have definitely I will update it base on the need.

hoektoe:
Can't wait for this to use with storyblok and nuxtjs/image

hoektoe:
Is it just the merge conflict needing to be resolved or is there other issues related to this PR for it to get merged into master?

f3ltron:
I am currently working on it @hoektoe. Thanks for your message :)

I need to validate with storyblok team first.

src/runtime/providers/storyblok.ts
```diff
@@ -0,0 +1,54 @@
...
+  return ret
+}
+
+export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = 'https://img2.storyblok.com' } = {}) => {
+  const mergeModifiers = overrideUndefined({
+    ...defaultModifiers,
+    ...modifiers
+  })
+
+  const srcWithoutBase = src.replace('https://a.storyblok.com', '')
```
pi0, line 48: We can use `withoutBase(url, base)` from ufo to remove base and `withBase (url, base)` to add baseURL.

src/runtime/providers/storyblok.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { width, height } = mergeModifiers
+
+  if (width === '0' && height === '0') { return '' }
+
+  return `/${width}x${height}`
+}
+
+const buildFit = ({ fit }) => fit ? `/fit-${fit}` : ''
+
+const buildSmart = ({ smart }) => typeof smart !== 'undefined' ? '/smart' : ''
```
pi0, line 18: Is there any reason not checking falsy only? (smart ? '/smart' : '')

src/runtime/providers/storyblok.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const defaultModifiers = {
+  width: '0',
+  height: '0'
+}
+
+const buildWidth = (mergeModifiers) => {
+  const { width, height } = mergeModifiers
+
+  if (width === '0' && height === '0') { return '' }
```
pi0, line 11: Can we check (if (!width && !height)) and not default them to 0?

hoektoe:
> I am currently working on it @hoektoe. Thanks for your message :)
> 
> I need to validate with storyblok team first.

Awesome, it's just a great integration. Enjoying both nuxtjs and storyblok together.

f3ltron:
Then even after I asked I have no update on it from Storyblok. A bit disappointed :(

pi0:
Thanks @f3ltron <3

f3ltron:
Thanks for your refactor :). I spoke with christian dev at storyblok they have validated it 👍🏼

christianzoppi:
Hi everyone 😄  I double-checked all the features of the Storyblok Image Service and they are working perfectly. Thanks, @f3ltron! 🙌🏻

PR #136, opened 2021-01-04
fix(payload): fixed payload import for lazy hydrated components

StephanGerbeth:
Currently the $nuxt._pagePayload is accessed directly on the client side. However, _pagePayload is overwritten by payload prefetches (resourceHints). This prevents lazy hydrated components from accessing the value.

Solution:
When a route is changed, the payload is written away via beforeNuxtRender on server side / nuxt-import on client-side and can be read by lazy hydrated components at a later time.


src/runtime/image.ts
```diff
@@ -106,6 +114,22 @@ async function getMeta (ctx: ImageCTX, input: string, options: ImageOptions) {
   return meta
 }
 
+async function fetchPayload (page, context) {
+  if (!context.nuxtState) {
```
farnabaz, line 118: The page payload uses on client side in full static websites, And fetching payload on the server side is meaningless. We could simply check for existence of `window.__NUXT_IMPORT__` function and import payload

src/runtime/image.ts
```diff
@@ -106,6 +114,22 @@ async function getMeta (ctx: ImageCTX, input: string, options: ImageOptions) {
...
+    context.beforeNuxtRender(({ nuxtState }) => {
+      pagePayload = nuxtState.data[0]
+    })
+  } else {
+    try {
+      // @ts-ignore
+      const payload = await window.__NUXT_IMPORT__(decodeURI(page.path.replace(/\/$/, '')), encodeURI(context.nuxtState.staticAssetsBase + page.path + 'payload.js'))
```
farnabaz, line 125: We should use `getStaticAssetsPath` for generating payload path, `getStaticAssetsPath` will handle the router base and generates correct path.
Checkout https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/App.js#L313

farnabaz (changes requested):

src/runtime/image.ts
StephanGerbeth: should this be changed?

src/runtime/image.ts
StephanGerbeth: done

pi0:
Hi! State hydration has been rewritten and simplified ([code when writting this comment](https://github.com/nuxt/image/blob/57fd139f6c2e0cb16d37c1de1c7346c5f5596ae1/src/runtime/image.ts#L26)). I hope this fixes this issue with 0.2.x. Otherwise please ping or make an issue to investigate.

Issue #148, opened 2021-01-14
Unable to set fallback image & failed, reason: socket hang up


treboryx:
Unlike the <img> tag, I'm unable to use @error and use a method to set a fallback image in case the image that's loaded is broken, is there a way of doing this?

Also, on my development environment I often get the following error

```
failed, url-of-image-here reason: socket hang up
```

Which doesn't seem to be the case in a production environment.

pi0:
Hi @treboryx. Can you please try behavior with `0.2.0-alpha.7`? We no longer fetch placeholder images

treboryx:
> 0.2.0-alpha.

new error occurred

```
[Vue warn]: Unknown custom element: <nuxt-image> - did you register the component correctly? For recursive components, make sure to provide the "name" option.
```

is there something that I'm missing? With the 0.0.4 version this error doesn't not appear.

pi0:
You need to change `nuxt-image` to either `nuxt-picture` or `nuxt-img`. 

Sorry for this docs are still outdated. And image-module is marked as WIP because api changes can happen until we finally announce module.

treboryx:
> You need to change `nuxt-image` to either `nuxt-picture` or `nuxt-img`.
> 
> Sorry for this docs are still outdated. And image-module is marked as WIP because api changes can happen until we finally announce module.

Yes that seems to work. However, I must mention that I no longer see that smooth lazy loading animation anymore. But like you said it's WIP.

We can wrap this up once you review PR https://github.com/nuxt/image/pull/153 since it's part of the second issue I mentioned here.

pi0:
Closing this as hang-up is resolved. We can follow feature request in #153 :)

PR #153, opened 2021-01-21
feat(nuxt-img): support fallback prop

treboryx:
issue was part of https://github.com/nuxt/image/issues/148 that I created myself, thought I'd code it myself into nuxt-image


pi0:
This would be a nice feature. I'm just wondering what happens if fallback image also makes an error

treboryx:
> This would be a nice feature. I'm just wondering what happens if fallback image also makes an error

Fallback image would be something that would always exist, like a local static file image. I don't think there's anything we can do if fallback image errors, but I can't really see why that would happen.

pi0:
By correct usage it should be totally fine but if wrongly using a remote image which cannot be loaded (404 or network error) users might end with an infinitie loop which we should consider when adding feature

treboryx:
> By correct usage it should be totally fine but if wrongly using a remote image which cannot be loaded (404 or network error) users might end with an infinitie loop which we should consider when adding feature

If it's a local file you don't run into the infinite loop. For remote images we can use node-fetch, which I noticed is already a dependency, and if the http status is something other than 200, return or set it to an empty image (a base64 one, similar to the GIF one you already have).


Sounds good?

pi0:
@treboryx I think simply keeping a flag like `this._errored` is enough to set fallback twice and avoid loop

Also would be nice if you remove `package.lock.json` from PR

treboryx:
> @treboryx I think simply keeping a flag like `this._errored` is enough to set fallback twice and avoid loop
> 
> Also would be nice if you remove `package.lock.json` from PR

Yeah that's a better idea, all done, works like a charm!

pi0 (approved):

Atinux (approved):

Atinux:
One note, shall we also `$emit('error', event)` in `setFallbackImage` ?

treboryx:
> One note, shall we also $emit('error', event) in setFallbackImage ?

A 404 not found error is already logged in the console, I don't think this is necessary.

pi0:
> One note, shall we also $emit('error', event) in setFallbackImage ?

Makes sense since when fallback prop is not set, we expect identical behavior of img and nuxt-img for custom error handling, I will take care before merge :)

treboryx:
Hi, do we know when is this going to be merged into an update? 👉👈🥺

treboryx:
Hi again. Is there any chance you know when this is going to be merged & released with a new update?

pi0:
Hi @treboryx. I'm SO sorry that your PR is stalled for long time. This is mostly pending because we need consistent support with `<nuxt-picture>` and also might be probably better supporting fallback in placeholder component (#189) that can properly control error handling behavior (rather than directly updating `el` reference).

pi0:
In the meantime you can create a wrapper component. See this sandbox: https://codesandbox.io/s/trusting-euler-8h0j9?file=/components/nuxt-img-container.vue:0-482

```vue
<template>
  <nuxt-img v-bind="$attrs" @error.native="setFallbackImage" />
</template>

<script>
export default {
  props: {
    fallback: { type: String, required: false, default: undefined },
  },
  data() {
    return {
      hasError: false,
    };
  },
  methods: {
    setFallbackImage(event) {
      if (this.hasError) {
        return;
      }
      this.hasError = true;
      if (this.fallback) {
        event.target.src = this.fallback;
      }
    },
  },
};
</script>
```

treboryx:
> In the meantime you can create a wrapper component. See this sandbox: https://codesandbox.io/s/trusting-euler-8h0j9?file=/components/nuxt-img-container.vue:0-482
> 
> ```vue
> <template>
>   <nuxt-img v-bind="$attrs" @error.native="setFallbackImage" />
> </template>
> 
> <script>
> export default {
>   props: {
>     fallback: { type: String, required: false, default: undefined },
>   },
>   data() {
>     return {
>       hasError: false,
>     };
>   },
>   methods: {
>     setFallbackImage(event) {
>       if (this.hasError) {
>         return;
>       }
>       this.hasError = true;
>       if (this.fallback) {
>         event.target.src = this.fallback;
>       }
>     },
>   },
> };
> </script>
> ```

Thanks. Looking forward to official support.

PR #154, opened 2021-01-22
feat: support modifiers prop for components

f3ltron:
I have a concern about modifiers.

Every providers can have really differents modifiers from one to another. Would not be better to have like:
- Each provider receive one option(s) object where are define all available modifiers
- He is doing its own check / test
- So with that you can delete the default one. 

I know it will not be like on props for fit etc but one object. In my opinion It will be clearer and more modular.

What do you think ?

 @pi0 what do you think?


src/runtime/components/nuxt-img.vue
```diff
@@ -60,7 +61,8 @@ export default {
         format: this.format,
         quality: this.quality,
         background: this.background,
-        fit: this.fit
+        fit: this.fit,
+        options: this.options
```
pi0, line 65: We do not support `modifiers.options` it should be `...this.modifiers` i guess

src/runtime/components/nuxt-picture.vue
```diff
@@ -63,6 +63,7 @@ export default {
     quality: { type: [Number, String], required: false, default: undefined },
     background: { type: String, required: false, default: undefined },
     fit: { type: String, required: false, default: undefined },
+    options: { type: Object, required: false, default: () => {} },
```
pi0, line 66: Default is not needed

src/runtime/components/nuxt-picture.vue
```diff
@@ -120,6 +121,7 @@ export default {
         format: this.format,
         quality: this.quality,
         background: this.background,
+        options: this.options,
```
pi0, line 124: Same for `...this.modifiers` (and probably should be first)

src/runtime/components/nuxt-img.vue
f3ltron: ...this.options you mean ?

src/runtime/components/nuxt-picture.vue
f3ltron: you mean ...this.options i guess and yea agreed

pi0 (approved):

PR #162, opened 2021-01-26
fix(plugin): create static manifest for client hydration

farnabaz:
This PR will create a manifest to hold static image mapping and fix lazy hydration issue.  
The issue is introduce in #136 by @StephanGerbeth.


pi0:
Context: There is a known issue `__NUXT_IMPORT__ ` is not defined with nuxt-edge

Issue #161, opened 2021-01-26
some bugs in custom provider (default provider, shortcut, types)


reslear:
` "nuxt-edge": "2.15.0-26856070.8350560"`
`"@nuxt/image": "^0.2.1"`

iam search bugs:
1. nuxt-picture `provider` and default custom provider not working (nuxt-img work)
 https://github.com/nuxt/image/blob/d2a9cdd153dc77d5245a58e0f49caebc4cd14662/src/runtime/components/nuxt-img.vue#L54-L58 don't find in https://github.com/nuxt/image/blob/d2a9cdd153dc77d5245a58e0f49caebc4cd14662/src/runtime/components/nuxt-picture.vue#L194-L203
2. work in nuxt-config path only dot start `./providers/custom` not shortcut `~/providers/custom`
3. not importable types for custom provider `getImage`   - `import type {} from "@nuxt/image"`
missing in dist types this https://github.com/nuxt/image/blob/d2a9cdd153dc77d5245a58e0f49caebc4cd14662/src/types/image.d.ts#L19

pi0:
Hi @reslear thanks for making issue. Please note that image module is WIP both API and Docs so anything can be expected. I will keep it open until fixing.

reslear:
@pi0 yes i know that this is WIP) this is my little help, if you can call it that) maybe next my step opening PRs, thanks)

pi0:
1 should be fixed by #169, 2 by 930e192 and for 3 feel free making a PR (couldn't follow issue since `ProviderGetImage ` is already exerted)

reslear:
hi @pi0 i am testing new version 0.3, but `ProviderGetImage` not include `export type` in dist `types.d.ts`
for importing  

<img width="1132" alt="Снимок экрана 2021-02-04 в 12 58 02" src="https://user-images.githubusercontent.com/12596485/106877233-96e5fc80-66e9-11eb-813e-2ad4f2925f3f.png">


```ts 
import {ProviderGetImage} from "@nuxt/image"
export const getImage: ProviderGetImage = () =>  
```

it seems `siroc` merge all types and cut public exports

reslear:
hi @danielroe https://github.com/nuxt-contrib/siroc/commit/ebc5a02f14ab962666ae2d3e784f188c01677331  will solve this problem https://github.com/nuxt/image/issues/161#issuecomment-773210145  ?

danielroe:
Yes, this appears to be fixed in the latest version of `siroc`.

[...]

PR #171, opened 2021-01-27
chore(playground): add custom provider

reslear:
#161


pi0:
Hi @reslear thanks for PR! For making custom providers with typescript, we may import `ProviderGetImage`  from '@nuxt/image'. NuxtAppOptions should be only extended by what module plugins _inject_ into context (like `$img` or `$axios`)

PR #182, opened 2021-02-08
feat!: screen based responsive

pi0:
Currently, responsive prop is generating breakpoints by assuming image layout is fuild (filling screen width).

- Global `sizes` option replaced by `screens`
- Remove `responsive` attr from `nuxt-img`
- `sizes` attr for `nuxt-picture` and `nuxt-img` accept a new object
- Rename `srcSet` to `srcset` everywhere
- Computed value for `srcset` and `sizes` improved

Usage with `<nuxt-img>`:

```vue
<NuxtImg sizes="sm:100vw md:50vw lg:400px" />
```

Generates:

```html
<img src="https://res.cloudinary.com/nuxt/image/upload/w_500/remote/nuxt-org/blog/going-full-static/main" alt="main" sizes="(max-width: 640px) 100px, 500px" srcset="https://res.cloudinary.com/nuxt/image/upload/w_100/remote/nuxt-org/blog/going-full-static/main 100w, https://res.cloudinary.com/nuxt/image/upload/w_500/remote/nuxt-org/blog/going-full-static/main 500w">
```

Default value for screens: (inspired by tailwind [breakpoints](https://tailwindcss.com/docs/breakpoints))

```js
{
    screens: {
      xs: 320,
      sm: 640,
      md: 768,
      lg: 1024,
      xl: 1280,
      xxl: 1536,
      '2xl': 1536
    },
}
```

### Try locally

- Checkout branch
- `yarn && yarn dev`
- http://localhost:3000/responsive


pi0:
@Atinux Both good ideas. Just consider for fluid support we need: 

1. A way to specify we want fuild behavior (by either `flux` prop or using string for one of the sizes)
2. A way to specify at which screen breakpoint (max-width as number or by name like `md`) we want to swtich to fluid layout
3. A way to specify `%vw` relation between fluid image and screen size (like 100 or 80)
4. Also as mentioned regarding tricky part, there might be more than one fluid breakpoint (tablet, desktop or orientation based) so we would need multiple values for `fluid`.

I would suggest an API like below:

```vue
<nuxt-img :fluid="{ lg: 60, md: 80, sm: 100 }" />
```
(`fluid` covers (1), `md` covers (2), `80` covers (3) and comma seperated values cover (4))

Generates:

```html
<img src="damavand.jpg?w=800" 
sizes="(max-width: 640px) 100vw, (max-width: 768px) 80vw, (max-width: 1024px) 60vw"
srcset="img.jpg?w=640 640w, img.jpg?w=614 614w"
>
```

Remarks:

-  sm = 640
-  md = 768
-  lg = 1024
- 640  * 1.0 = 640
-  768  * 0.8 = 614
-  1024 * 0.6 = 614

`614w` variant is used only for fluid size of both md and lg since for `lg` screens we are using 60vw which is almost same as 80vw of `md`

Atinux (approved):

PR #195, opened 2021-02-18
feat: expose vetur helpers

reslear:


reslear:
hi @pi `mkdist` or maybe use native `cp -r` ?)

pi0:
Hi @reslear and thanks for PR. I think for time being we can use top-level vetur since it is not auto-generated.

PR #199, opened 2021-03-04
fix(observer): add `mediaQueryList` listener fallback for safari/older browsers

danny-pearson:
Not sure if this is already in the process of being done, but a fallback is needed for the print listener on Safari <=13 (not sure on 14) as `MediaQueryList` still only supports `addListener` :)


pi0 (approved):
Thanks!

PR #205, opened 2021-03-27
feat: Adds logic to enable secure URLs in the imgix provider

danny-pearson:
This enables the Imgix provider to sign the image url using the token provided when 'Secure URLs' is enabled in the source. The token can be passed in as an option to the provider in `nuxt.config`...

Screenshot 2021-03-27 at 19 07 35

This also lets you use images that are served from a web proxy source without having to sign them manually or rely on another dependency :)


pi0:
Hi @danny-pearson thanks for the PR. I think we need a mechanism like a runtime config to pass the token. Since options are directly added to runtime/providers which is bundled as a part of client-code (this means someone can easily gain token to sign their URLs)

danny-pearson:
@pi0 Aye I'm not ganna lie that completely slipped my mind baha :/

I'll have a look at getting that sorted this weekend :)

pi0:
We may use new [provider setup](https://github.com/nuxt/image/blob/082ae1f8487c45700ee941a26e18a29b9faf5e8c/src/provider.ts#L18) hooks to inject token to `nuxt.options.privateRuntimeConfig` (it won't work for client-generated URLs)

pi0:
I've moved this topic to #276. Thanks for initiative @danny-pearson 🙏

PR #206, opened 2021-03-30
Added nuxt-download-link component to generate and crawl image/PDF download links

ambroisemaupate:
When generating static website with content and image coming from a temporary API, we need to crawl and store images/PDF files in `_nuxt` folder in the same way *nuxt/image* crawl and store image files locally at `nuxt:generate` stage.

This PR adds a new `nuxt-download-link` component to build a `<a download>` element following the same logic as `` and `<picture>`

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -71,7 +71,7 @@ async function imageModule (moduleOptions: ModuleOptions) {
   })
 
   nuxt.options.build.loaders = defu({
-    vue: { transformAssetUrls: { 'nuxt-img': 'src', 'nuxt-picture': 'src' } }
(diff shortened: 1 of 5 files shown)


pi0:
Hi @ambroisemaupate thanks for PR 🙏 And sorry for the late review. 

I think what you are trying to achieve, is simply possible using [$img utility](https://image.nuxtjs.org/api/$img). This will also work with crawler/static target.

```html
<a :to="$img('/images/test.jpg', { width: 120 })" />
```

(alternatively you can create higher order component in `components/` directory)

I'm closing this since it is easily possible and helps to keep the image module simple yet flexible enough for specific use cases like download links.

Still, if you believe this pattern is common enough to have direct module support, feel free using an Issue to discuss.

PR #207, opened 2021-03-30
Expose nuxtImage:updateImageMap event globally

ambroisemaupate:
- **Target**: static 
- **Current behaviour**: `staticImageMap` is updated only `afterEach` router changes.

Expose `nuxtImage:updateImageMap` event globally to be able to force *nuxt/image* to append `_pagePayload?.data?.[0]?._img` from any other generated payload without any Route change or on `beforeRouteUpdate`.

```typescript
beforeRouteUpdate(to: Route, from: Route, next: Function) {
    if (this.$store.state.isAFancyEmbeddedPage) {
        /* new data, new components requiring some nuxt-img or nuxt-picture */
        this.pageData = this.$store.state.fancyEmbeddedPageData
        // Need to emit an event to force NuxtImage to append new pagePayload._img
        // to current page staticImageMap
        this.$nuxt.$emit('nuxtImage:updateImageMap')
    } else {
        next()
    }
}
```

### DIFF
--- a/src/runtime/utils/static-map.ts
+++ b/src/runtime/utils/static-map.ts
@@ -18,7 +18,11 @@ export function useStaticImageMap (nuxtContext?) {
 
   // Make sure manifest is initialized
   if ((window as any).onNuxtReady) {
-    (window as any).onNuxtReady(updateImageMap)
+    (window as any).onNuxtReady(function () {
+      updateImageMap()
+
+      window.$nuxt.$on('nuxtImage:updateImageMap', updateImageMap)
+    })
   }
 
   return staticImageMap


pi0:
Hi @ambroisemaupate thanks for PR. Would you please create a reproduction for current issue? So we can discover possible solutions for this.

manuelodelain:
Hi @pi0 

Here is a reproduction of the use case https://github.com/manuelodelain/nuxt-image-update-image-map

The scenario is: we have to open a page into a modal, but keep the ability to have this same page as a regular page (ie with a proper url).
The idea is to keep the lifecycle of a regular page (fetch the data with `asyncData()`), and if the page has to be opened as a modal the router is stopped, we keep the data from the `asyncData()`, inject it into a modal component and we handle manually the url with the History API.

- target: static
- Expected behavior: After the generation, the images should be served from the `dist/` folder
- Current behavior: After the generation, if the router is interrupted, the images use the original src attribute.

The issue here is that Nuxt Image updates the static image map using the `afterEach()` guard. 
This PR exposes a global event to manually update the static image map if needed.

ambroisemaupate:
Hi @pi0 

Have you any feedback on this PR? 

Thanks

pi0:
Hi @ambroisemaupate. Can you still reproduce it on ^0.4.14?

ambroisemaupate:
> Hi @ambroisemaupate. Can you still reproduce it on ^0.4.14?

Yes

danielroe:

PR #217, opened 2021-04-18
fix: limit static download concurrency

jdempster:
Resolve error 

ERROR  request to http://localhost:57016/_ipx/example.webp?w=768&f=webp failed, reason: connect ECONNRESET 127.0.0.1:57016                 @nuxt/image 18:04:53

Happens during generate when there are a lot of images in your project, they all make requests to ipx.
The fix here uses pLimit to limit the number of concurrent requests to 5.


src/generate.ts
```diff
@@ -27,15 +28,16 @@ export function setupStaticGeneration (nuxt: any, options: ModuleOptions) {
 
   nuxt.hook('generate:done', async () => {
     const { dir: generateDir } = nuxt.options.generate
+    const limit = pLimit(5)
```
pi0, line 31: It is not clear to me without reproduction at what concurrency issue happens but I think we should be safe setting it to 8 or 16.

pi0 (approved):

pi0:
Thanks!

Skrapion:
I'm using this fix successfully on 0.0.4 (until placeholders come back!). Anyone who needs it can update their package.json to:
```
"dependencies": {
        "@nuxt/image": "github:skrapion/image#oldfix"
}
```

PR #218, opened 2021-04-19
feat: support cloudinary fetch base url

harlan-zw:


src/runtime/providers/cloudinary.ts
```diff
@@ -3,6 +3,8 @@ import { joinURL } from 'ufo'
 import { createOperationsGenerator } from '~image'
 
 const convertHextoRGBFormat = (value: string) => value.startsWith('#') ? value.replace('#', 'rgb_') : value
+const removePathExtension = (value: string) => value.replace(/\.[^/.]+$/, '')
+const removePathLeadingSlash = (value: string) => value.replace(/\/$/, '')
```
pi0, line 7: We can use `withoutLeadingSlash` from `'ufo'`

harlan-zw: Thanks for the quick review @pi0. Realised I had named that function incorrectly as well, it's the trailing slash we want to remove so both `/fetch/` and `/fetch` would be supported as base URLs.

Updated

pi0 (approved):

pi0:
Thanks :)

PR #222, opened 2021-04-20
feat: type improvements

danielroe:
Briefly:
* move whole project to strict typing
* fully type mixins/components
* address any issues flagged in the process


src/runtime/components/nuxt-img.vue
```diff
@@ -34,7 +42,8 @@ export default {
       }
       return this.$img(this.src, this.nModifiers, this.nOptions)
     },
-    nSizes () {
+    /* eslint-disable no-undef */
+    nSizes (): ReturnType<typeof $Img['getSizes']> {
```
danielroe, line 46: We could also import the type directly. It will be excluded either way. Thoughts?

pi0: I think while we are moving towards strict typing, makes sense to also care about naming internal interfaces

src/runtime/components/image.mixin.ts
```diff
@@ -25,41 +30,52 @@ export const imageMixin = {
...
         longdesc: this.longdesc,
         ismap: this.ismap,
-        crossorigin: this.crossorigin,
+        crossorigin: this.crossorigin === true ? 'anonymous' : this.crossorigin || undefined,
         loading: this.loading,
         decoding: this.decoding
       }
     },
-    nModifiers () {
+    nModifiers (): { width?: number, height?: number, format?: string, quality?: string | number, background?: string, fit?: string } & Record<string, any> {
```
pi0, line 63: This return type (and rest of the functions) are already inferred by typescript. (also mkdist at least currently for some reason is not preserving types at all)

src/runtime/components/nuxt-img.vue
pi0: 892b3f9509b80727803a995dec282c62a6d2f511

pi0 (approved):

Issue #216, opened 2021-04-17
Responsive image sizing


ArtCraft:
We live in a time when websites should work on a huge range of devices with different sizes and densities, it is not enough to have only one version of image... at least 2 versions of each file (1x and 2x) is a must for modern web, but in many cases it may be reasonable to have 3-4 sizes of each file in webp format + fallback to jpg/png, so each picture usually may be resized/converted to 2-4-6 or more versions.

Problems:
- There is no way do define image sizes for high density image versions
- Presets do not allow to use `sizes` property available for `nuxt-img`
-  (for @nuxt/content) Images inside .md files that included using .md syntax are not processed by @nuxt/image, it would be very nice if it is possible to set a default sizing using preset for those images

ArtCraft:
for example
```
<nuxt-img src="/picture1.jpg" sizes="sm:200px, md:500px"></nuxt-img>
```
this generates following code
```
<img
    srcset="/_ipx/pic1.jpg?w=200 200w, /_ipx/pic1.jpg?w=500 500w"
    sizes="(max-width: 640px) 200px, 500px"  
    src="/_ipx/pic1.jpg?w=640"    
 >
```

on retina display (device pixel ratio 2.0) optimal image sizes would be 400px and 1000px, but srcset attribute do not provide those sizes

what should be generated to look good is:
```
<img
          srcset="/_ipx/pic1.jpg?w=200 200w, /_ipx/pic1.jpg?w=400 400w, /_ipx/pic1.jpg?w=500 500w, /_ipx/pic1.jpg?w=1000 1000w"
          sizes="(max-width: 800px) 200px, 500px"
          src="/_ipx/pic1.jpg?w=50" alt="test"
        >
```

so there should be some way to define those sizes as well
something like

```
<nuxt-img src="/picture1.jpg" sizes="sm:200px+2x, md:500px+2x"></nuxt-img>
or
<nuxt-img src="/picture1.jpg" sizes="sm:200px, md:500px" resolutions="1x, 2x"></nuxt-img>
```

pi0:
Hi @ArtCraft. Thanks for detailed proposal. Supporting densities, for sure makes sense. But I am thinking if main purpose is that, we can think about easier interface like this:

```html
<nuxt-img sizes="sm:200px, md:500px" densities="1x 2x">
```

[...]

PR #227, opened 2021-04-23
Feat: Add srcset property to nuxt-img and nuxt-picture

dannyalder88:
As suggested in #216 the current implementation has poor support for handling different device pixel densities. This adds the _srcset_ prop as suggested by ArtCraft so you can control which image sizes are created. It supports a string of widths in pixels to overwrite the images generated by the _sizes_ prop or pixel density descriptors to adjust which images the _sizes_ prop generates.

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -194,17 +194,27 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
   // string => array
   if (typeof opts.srcset === 'string') {
     for (const entry of opts.srcset.split(/[ ,]+/).filter(e => e)) {
-      const size = parseInt(entry)
-      if (isNaN(size)) {
-        continue
+      if (/^[x]\d+/.test(entry)) {
+        const dpiSize = parseInt(entry.replace(/^x/, ''))
+        if (isNaN(dpiSize)) {
+          continue
+        }
+        srcset.push(...sizeVarients.map(s => s.width * dpiSize))
+      } else {
+        const size = parseInt(entry)
+        if (isNaN(size)) {
+          continue
+        }
+        srcset.push(size)
       }
-      srcset.push(size)
     }
   } else if (opts.srcset) {
     srcset.push(...opts.srcset.filter(e => !isNaN(parseInt(e))))
   }
 
-  srcset.forEach((width) => {
+  const uniqueSrcset = Array.from(new Set(srcset))
+
+  uniqueSrcset.forEach((width) => {
     const height = ratio ? Math.round(width * ratio) : parseSize(opts.modifiers.height)
     srcVarients.push({
       width,
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -191,7 +191,7 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
     defaultVar.media = ''
   }
 
-  // string => array
+  // Only use srcset if it is a string
   if (typeof opts.srcset === 'string') {
     for (const entry of opts.srcset.split(/[ ,]+/).filter(e => e)) {
       if (/^[x]\d+/.test(entry)) {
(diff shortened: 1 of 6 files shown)


src/runtime/image.ts
```diff
@@ -173,7 +175,7 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
       width = Math.round((width / 100) * screenMaxWidth)
     }
     const height = ratio ? Math.round(width * ratio) : parseSize(opts.modifiers.height)
-    variants.push({
+    sizeVarients.push({
```
hya, line 178: spelling: sizeVariants

src/runtime/image.ts
```diff
@@ -140,9 +140,11 @@ function getPreset (ctx: ImageCTX, name?: string): ImageOptions {
 
 function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
   const ratio = parseSize(opts.modifiers.height) / parseSize(opts.modifiers.width)
-  const variants = []
+  const sizeVarients = []
+  const srcVarients = []
```
hya, line 144: Spelling - Variants

src/runtime/image.ts
```diff
@@ -182,16 +184,47 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
...
-  const defaultVar = variants[variants.length - 1]
+  const defaultVar = sizeVarients[sizeVarients.length - 1]
   if (defaultVar) {
     defaultVar.media = ''
   }
 
+  // Only use srcset if it is a string
+  if (typeof opts.srcset === 'string') {
+    for (const entry of opts.srcset.split(/[ ,]+/).filter(e => e)) {
+      if (/^[x]\d+/.test(entry)) {
```
hya, line 197: The regex matches on x9999999 - maybe make it so that it rejects anything more than a single digit?

dannyalder88: Sorry, took a while to get around to this. The regex now checks to make sure that only a single digit is used in the descriptor

pi0:
Hi @dannyalder88 Thanks for PR (and fixing typos). If you have a chance to check my comment https://github.com/nuxt/image/issues/216#issuecomment-840740583, if pure goal is supporting densities, I think it might be much easier for both users and maintenance to keep it as simple with `sizes` property (to be supported in `nuxt-img` as well).

src/runtime/image.ts
hya: Thanks for making the changes!

dannyalder88:
Finally got round to implementing the changes discussed in #216 (sorry it took a while been a busy month).

The latest changes change the srcset prop to a densities prop that only accepts single-digit density descriptions (1x 2x etc.).
I've also added a global densities option that can be set in nuxt.config to make it easier to have all images produce the users preferred densities.

I've noticed there's a lot more typescript being used throughout the project now. I'm not that familiar with it so if I've done anything wrong let me know and I can get it changed.

madsh93:
Is it correctly understood that this will enable e.g. retina images?

<nuxt-img width="500" height="200" densities="1x 2x">

which will output images with:
width = 500 and height = 200 if devices pixel density is 1x
and
width = 1000 and height = 400 if pixel density is 2x

dannyalder88:
Yes, it would generate retina sizes. It uses the size prop to create the files. So it will take what nuxt/image calculates as a size and then will use the density prop to generate the additional image sizes for retina. So if the sizes passed were "sm:100, md:400" this would generate images of 100, 200, 400 and 800 pixels wide. It adds these sizes to the generated code to let the browser know they are available and select the best image for the displayed width and pixel density.

tetz2442:
Is there any progress on this? I'd love to get this on my blog to better handle retina screens

adinvadim:
Hi! What is left to fix so that you can merge?

dannyalder88:
Hey! Sorry, I've been absent on this. This is the first PR I've done for a package so I'm not 100% sure about the process of getting it merged/approved. I was just under the assumption it had been decided not to merge the feature for now as a package maintainer has to do it.

If there is still anything outstanding that I need to do my side let me know and I'll get it sorted.

itpropro:
@pi0 has this been superseded by #459 and can be closed?

zainspresso (approved):

IBM51OO (approved):

danielroe:

Issue #215, opened 2021-04-16
Nuxt Generate does not produce images


dannyalder88:
**Nuxt version: 2.15.3
@nuxt/image version: 0.4.4
target: 'static'**

I have been using the module in dev mode with no problems. However, when trying to generate a static site using `nuxt generate` no images are produced by @nuxt/image and all images on the site 404. The generated image URLs point to _ipx/IMAGE_PATH but this is not in the dist folder of the Nuxt project.

I have tried with a fresh Nuxt project with only @nuxt/image added and get exactly the same issue. Running `nuxt generate` produces no errors in the CLI and the site builds without a problem. The images are just not generated.

Any help on this would be greatly appreciated.

tusamni:
Having a similar issue publishing to Vercel.

dannyalder88:
@farnabaz Replicating my code on template.nuxtjs.org resulted in the module working how it should. After going through the code and trying to spot differences it looks like having `ssr: false` set in the nuxt.config.js was preventing nuxt/image from running correctly during `yarn generate`. When running `yarn create nuxt-app <project-name>` and selecting static as the target mode, this flag is added to the config by default. If it's not a bug, this might need a warning added in the documentation for the module as I couldn't find information about it currently.

ririen:
I confirm the same bug on Vercel.

tusamni:
To be fair, I can't get it to work on "npm run generate" either, which is what Vercel is using to generate the site.

ririen:
There is probably something wrong with the package here. At least I think so. I have tried 2 versions.

Vercel based on: https://github.com/nuxt/nitro-demo
Vercel based on: https://github.com/nuxt/vercel-builder

Neither the generated version nor the server version works with displaying images using this package. 

https://github.com/nuxt/vercel-builder/issues/548 ping @pi0 You have created this integration. Do you have an idea what could be wrong?

tusamni:
For what it's worth I was able to get it working with the following config: 
`    buildModules: [
        // Next Image module https://image.nuxtjs.org/components/nuxt-img
        ['@nuxt/image', {
            provider: 'static',
            dir: "assets/images",
        }],
    ],`

tanc:
The key for me was `provider: 'static'` and then images got generated, even from remote sources.

shadow81627:
[...]

Issue #167, opened 2021-01-27
Image URL in srcset not set correctly


thmsnhl:
I am using `nuxt-img` in a project with deployment target "static", images are pulled from a Content Management System and not locally hosted.
I do not specify a provider or define any other options in `nuxt.config.js`.

When running `nuxt generate` everything seems to be working as intended and the build logs output `✔ Generated static image dist/_nuxt/image/{HASHED-FILENAME}.jpg` for every image. But when I inspect the DOM, the created `<picture>`-Elements contain a lot of `<source>`-Elements with the initial image URL instead of the freshly generated one.

pi0:
This is a known issue (see #166)

thmsnhl:
@Atinux I am using the latest Version 0.4.0 

@pi0 Oh sorry, I didn't see this one when scrolling through the existing issues

thmsnhl:
@pi0 I just looked into #166 and I don't see a relation between the described issue there and mine. 

To clarify: the rendered URL in the `<source>`-Element is exactly the URL used in `<nuxt-img :src="${CMS_IMAGE_URL}" />`.

pi0:
Latest version is `0.2.1` (0.0.4 is old). Original URLs are being used as fallback when resized version is not included in payload mapping with default provider (`static`) (which addressed in #162 sorry i used wrong ref)

It would be nice if you can try it and if not working sharing your repository.

thmsnhl:
Oh sorry, I also had a typo in the version number. I use latest version which is indeed `0.2.1` 🤦 

I'll try to validate the payload on my own, unfortunately I am not allowed to share the repository.

thmsnhl:
So I took a look into `_nuxt/static/article/name-of-the-article/payload.js` and found the following:
```js
  ... 
  return {
    data: [
      {
        article: {...},
        _img: {
          '/_img/remote/_/w_640/https://the.api.url/assets/iStock-1146697288.jpg': 'e1474b.jpg',
          '/_img/remote/_/w_750/https://the.api.url/assets/iStock-1146697288.jpg': '7ee5ad.jpg',
          '/_img/remote/_/w_828/https://the.api.url/assets/iStock-1146697288.jpg': '624500.jpg',
          '/_img/remote/_/w_1080/https://the.api.url/assets/iStock-1146697288.jpg': 'a6ac68.jpg',
          '/_img/remote/_/w_1200/https://the.api.url/assets/iStock-1146697288.jpg': '524fa4.jpg',
[...]

PR #236, opened 2021-04-27
fix: ensure static image map is available on first load

### LINKED ISSUE #234: $img does not return the correct url on nuxt generated pages
The `$img` function correctly generates the image on `nuxt generate` and returns the url to the generated image, but when viewing the generated page the returned url is incorrect and the image does not load.

For example the following (simplified) code in a page component:

```vue
<template>
  <div class="page-wrapper">
    <SocialHead :image="getMetaImage" />
  </div>
</template>

<script>
export default {
  computed: {
    getMetaImage() {
      const imageSrc = this.$img('https://back.wildrockmtb.com/uploads/tanc_jump_51512c7cec.jpg', { width: 1000, height: 100, quality: 65 }, { provider: 'static' })
      console.log('Generated source: ', imageSrc)
      return imageSrc
    },
  },
}
</script>
```
When running `nuxt generate` the following is shown in the output:

```
Generated source: /_nuxt/image/9014ee.jpg
```

But when viewing the generated page, console displays:

```
Generated source:  https://back.wildrockmtb.com/uploads/tanc_jump_51512c7cec.jpg
```

This is also the url passed to the `:image` prop in the generated page, which is not the `$img` generated image.

How is it possible to obtain and use the correctly generated `/_nuxt/image/***` url when using the `$img` function?

danielroe:
* Worth also considering whether to use `fetch` (object syntax, but would require Nuxt 2.15+) vs `data` (?)

closes #234 

Likely also addresses https://github.com/nuxt/image/issues/167#issuecomment-826439905, https://github.com/nuxt/image/issues/215#issuecomment-826449532

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -40,8 +40,9 @@ export function createImage (globalOptions: CreateImageOptions, nuxtContext: any
 
       if (process.server) {
         const { ssrContext } = ctx.nuxtContext
+        const ssrState = ssrContext.nuxt
         const ssrData = ssrContext.nuxt.data[0]
-        const staticImages = ssrData._img = ssrData._img || {}
+        const staticImages = ssrState._img = ssrData._img = ssrData._img || {}
         const mapToStatic: MapToStatic = ssrContext.image?.mapToStatic
         if (typeof mapToStatic === 'function') {
           const mappedURL = mapToStatic(image)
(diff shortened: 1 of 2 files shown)


src/runtime/image.ts
```diff
@@ -40,8 +40,9 @@ export function createImage (globalOptions: CreateImageOptions, nuxtContext: any
 
       if (process.server) {
         const { ssrContext } = ctx.nuxtContext
+        const ssrState = ssrContext.nuxt
         const ssrData = ssrContext.nuxt.data[0]
```
pi0, line 44: we can refactor this 🙈

pi0 (approved):

tanc:
Confirmed that with these applied changes the client now gets the correct mapping returned. Thanks!

Issue #189, opened 2021-02-15
support placeholder


pi0:

Tragio:
Any ETA when the support for placeholder will come back? Really thank you for your awesome work!

shadow81627:
I use vuetify image component for the lazy loading source and generate a small 10 pixel image.

```html
<v-img
            :lazy-src="$img(src, { width: 10, quality: 70 })"
            :src="$img(src, { quality: 70, height: 500 })"
            height="500"
     </v-img>

```

creazy231:
also would really like to see such a feature 👍

agnelnieves:
same here

omp-git:
please merge this cool feature :)

mrleblanc101:
@pi0 Why is the default placeholder `10x10` ? 
Shouldn't it preserve the aspect-ratio of the original image ?
Currently it cause major layout shift.
I guess I could set the ratio using `:placeholder=[width, height]`, except that my Images come from `.yml` fetched by Nuxt-content so I don't know their dimension and they are not all the same.

Source image:
<img width="1912" alt="Capture d’écran, le 2022-07-15 à 00 01 38" src="https://user-images.githubusercontent.com/17392251/179147814-05c19aee-9866-46b1-9ef8-4bf8c1cc9da7.png">

Placeholder image:
<img width="1912" alt="Capture d’écran, le 2022-07-15 à 00 01 42" src="https://user-images.githubusercontent.com/17392251/179147845-f26eb8a3-1103-4ba6-a180-1429b61a7e7b.png">

mrleblanc101:
Otherwise it works well for static image.

Placeholder:
<img width="1912" alt="Capture d’écran, le 2022-07-15 à 00 05 16" src="https://user-images.githubusercontent.com/17392251/179148206-e4b0119b-b1f5-45e5-8ec4-ecef61b63d83.png">

Source:
<img width="1912" alt="Capture d’écran, le 2022-07-15 à 00 05 21" src="https://user-images.githubusercontent.com/17392251/179148220-b7544771-6306-4f81-a27b-00e3d9af43c5.png">

enricodeleo:
Is there any way I can tell placeholder to respect `preset` ?

Flo2315:
Hi @pi0 
Nice feature 🚀, but it doesn't seem to work with the cloudimage provider.

vedmant:
[...]

PR #240, opened 2021-04-28
feat: implement `<noscript>` wrapper for lazyloaded images

### LINKED ISSUE #50: Fast lightweight loading attribute polyfill
# Loading="lazy" is the future
I saw that internally you used the **IntersectionObserver API by default instead of the new loading attribute**.

## Compatibility
I know that the loading attribute is not well supported yet... but there's a lightweight polyfill that you could use today.
It is only 2.88 KB for the minified version. Moreover, the polyfill is compatible down to IE 9.

## The benefits
In order to make a huge improvement for todays web performance challenges, you could implement the loading attribute polyfill.
This means less code maintenance, consequently less potential technical debt and a faster seo friendly nuxt-image component.

It could also allow you to avoid development errors (and therefore new issues from the community) due to custom providers urls encoding, url transformations and static generation.

See: https://github.com/mfranzke/loading-attribute-polyfill

### LINKED ISSUE #190: prefer native lazy loading when possible

### LINKED ISSUE #202: support noscript

### LINKED ISSUE #213: SEO Problem: Images will never be indexed
I just found out that in the PageSource itself all images have this format:

``

That is a total disaster for all SEO Crawlers as they will never index any of the images on the site. Why not just display the image-url with` /_ipx/` in the front?

danielroe:
## supporting native lazy loading

> Ideally we'd like to support native lazy loading rather than instantiate an IntersectionObserver needlessly.

There are generally speaking two approaches:
1. **render an `` with a data attribute**. When the page starts loading (but before hydration), `src` can be swapped out with this data attribute if the browser supports lazy loading - but this means that non-JS web scrapers won't see the image, nor will users who have JS turned off in the browser. (Sample library implementing this approach: https://github.com/verlok/vanilla-lazyload).
2. **render `<noscript></noscript>`**. When the page starts loading (but before hydration), we swap this out on load for `` (with `src` if native lazyloading is supported and an empty gif if not). This ensures that search engines and users without JS can see the image. (Sample library implementing this approach: https://github.com/mfranzke/loading-attribute-polyfill)

A third approach:
* **render an `` with full `src` attribute**. When the page starts loading (but before hydration), we polyfill lazy loading via an injected script. However, if we load the script post-body, browsers will already have begun requesting the URIs for these images. If pre-body, we can't interact with the DOM to replace or modify these images.

It's complicated to render `<noscript>` with Vue given issues of inherited attributes (for example) so it needs to be injected with a render hook. We also inject a 598 byte minified script into the page body. On page load, but before hydration, this script swaps out the `<noscript>` for its `` contents (replacing the `src` with an empty gif if the browser doesn't support lazy loading). If the browser doesn't support lazy loading, then this script adds a 1.13kB lazy loading polyfill.

**TODO**:
- [ ] solution that doesn't require `@nuxt/image` to be added to `modules` for server-side rendering

**Resources**:
* https://web.dev/browser-level-image-lazy-loading/

closes #50, closes  #190, closes #202, closes #213


src/lazy-script.js
```diff
@@ -0,0 +1,24 @@
+const EMPTY_GIF = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7'
```
pi0, line 1: This belongs to runtime.

siroc.config.ts
```diff
@@ -0,0 +1,37 @@
...
+          }
+        ],
+        plugins: [
+          babel({
+            babelHelpers: 'inline',
+            presets: [
+              [
+                '@babel/preset-env', {
+                  targets: {
+                    ie: '9'
```
pi0, line 27: We might increase it since IntersectionObserver is never supported in IE. (safari 12 might be an option)

pi0:
The current approach is depending on dom modification in a pre-body script so why we can't take 3rd option?

pi0:
@danielroe As a suggestion to phase out complexities towards using native, I would suggest first drop the current polyfill system for `loading=lazy` then we can clearly make a workaround for safari that does not add the overhead of majority of chrome/ff/modern users.

danielroe:
@pi0 The current approach injects the script immediately before the closing body tag.

src/lazy-script.js
danielroe: Located outside purely for packaging reasons. (Both `mkdist` and `siroc` clear their target directories before build.)

pi0:
(PR pending for removing polyfill support first then we can properly benchmark for safari polyfill addition cost and method)

rollup.config.js
```diff
@@ -0,0 +1,40 @@
...
+      {
+        file: `./dist/runtime/scripts/${file}.min.js`,
+        format: 'iife',
+        plugins: [
+          terser()
+        ]
+      }
+    ],
+    plugins: [
+      babel({
```
pi0, line 26: Off-topic: But esbuild itself has the ability for transpiling 🙈

src/runtime/scripts/lazy-noscript.js
```diff
@@ -0,0 +1,31 @@
...
+
+noscripts.forEach(function (el) {
+  let html = el.innerHTML.trim().replace(/&lt;/g, '<').replace(/&gt;/g, '>')
+
+  if (!html.match(/data-src/)) {
+    return
+  }
+
+  if (!supportsLazyLoading) {
+    html = html.replace(/ src="([^"]*)"/, ' src="' + EMPTY_GIF + '"')
```
pi0, line 22:
```suggestion
    html = html.replace(/ src="([^"]*)"/g, ' src="' + EMPTY_GIF + '"')
```

pi0: We might convert `src` to `data-src` at same stage (and match by `loading=lazy`)

package.json
```diff
@@ -12,7 +12,7 @@
     "vetur"
   ],
   "scripts": {
-    "build": "siroc build && mkdist --src src/runtime --dist dist/runtime -d --ext js",
+    "build": "siroc build && mkdist --src src/runtime --dist dist/runtime -d --ext js && rollup -c",
```
pi0, line 15: Shall we use unbuild? :))

rollup.config.js
danielroe: Yes, now that we're targeting safari 12 the output is identical ... 👍

package.json
danielroe: Its time has come! 😁

danielroe: (I'm thinking using unbuild should be a separate PR though.)

pi0:
Closing in favor of #256 and adding back compat legacy support via placeholder component. (#189)

PR #245, opened 2021-04-29
docs: remove outdated alt tag (not currently produced)

danielroe:


pi0:
Alt tag is not computed but still a valid html attr :) (if stripped is a bug)

PR #246, opened 2021-04-29
fix(vercel): fix vercel auto-detection and provide local vercel mode

danielroe:
This fixes vercel auto-detection (note the issue was only with older deployments that didn't have system environment variables exposed to the build). 

In addition, if `vercel` provider is manually selected, in development `@nuxt/image` will provide warnings if un-configured widths are used (or if no width is set manually) to prevent issues in deployment. As `@nuxt/image` is transpiled, this will be tree-shaken out in build.

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -71,17 +71,22 @@ export function resolveProvider (nuxt: any, key: string, input: InputProvider): 
   }
 }
 
+const providerFlags = {
+  vercel: [process.env.VERCEL, process.env.NUXT_ENV_VERCEL_ENV, process.env.NOW_BUILDER].some(Boolean)
+}
+
 export function detectProvider (userInput?: string) {
   if (process.env.NUXT_IMAGE_PROVIDER) {
     return process.env.NUXT_IMAGE_PROVIDER
   }
 
   if (userInput && userInput !== 'auto') {
+    for (const [provider, enabled] of Object.entries(providerFlags)) {
+      if (userInput === provider && !enabled) {
+        return 'static'
+      }
+    }
     return userInput
-  }
-
-  if (process.env.VERCEL || process.env.NUXT_ENV_VERCEL_ENV) {
-    return 'vercel'
   }
 
   return 'static'
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -71,22 +71,17 @@ export function resolveProvider (nuxt: any, key: string, input: InputProvider): 
   }
 }
 
-const providerFlags = {
-  vercel: [process.env.VERCEL, process.env.NUXT_ENV_VERCEL_ENV, process.env.NOW_BUILDER].some(Boolean)
-}
-
 export function detectProvider (userInput?: string) {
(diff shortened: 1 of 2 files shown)


src/provider.ts
```diff
@@ -80,7 +80,7 @@ export function detectProvider (userInput?: string) {
     return userInput
   }
 
-  if (process.env.VERCEL) {
+  if (process.env.VERCEL || process.env.NUXT_ENV_VERCEL_ENV) {
```
pi0, line 83: Can you check if legacy `NOW_BUILDER` also exists? (in case nuxt template not detected)

danielroe: In fact, it's the only variable that seems to be set. 🙈

src/provider.ts
```diff
@@ -71,18 +71,23 @@ export function resolveProvider (nuxt: any, key: string, input: InputProvider):
...
   if (userInput && userInput !== 'auto') {
+    for (const [provider, enabled] of Object.entries(providerFlags)) {
+      if (userInput === provider && !enabled) {
+        return 'static'
+      }
+    }
     return userInput
   }
 
-  if (process.env.VERCEL) {
```
pi0, line 83: This regression needs to be fixed regardless and I think simply checking an inline env was enough. Users don't need to locally configure provider to `vercel` since it is supposed to auto-detected (thus default is `static` for local) We can introduce something like `devProvider` option to allow possibility of overriding for dev time and explicitly use a provider for prod

danielroe: @pi0 I've made a stab at a development mode for vercel but it sounds like you have something slightly different in mind - happy to revert and we can discuss better approach in separate PR if you think better.

src/runtime/providers/vercel.ts
```diff
@@ -3,7 +3,18 @@ import { stringifyQuery } from 'ufo'
 
 // https://vercel.com/docs/more/adding-your-framework#images
 
-export const getImage: ProviderGetImage = (src, { modifiers, baseURL = '/_vercel/image' } = {}) => {
+export const getImage: ProviderGetImage = (src, { modifiers, baseURL = '/_vercel/image' } = {}, ctx) => {
+  if (process.env.NODE_ENV === 'development') {
+    if (!modifiers?.width) {
+      console.warn(`A defined width must be provided to use the \`vercel\` provider. Warning originated from \`${src}\`.`)
+      return { url: '/_vercel/width-required' }
```
pi0, line 10: What about preserving original URL to avoid 404 (also in production)

src/provider.ts
pi0: This good track to me

src/runtime/providers/vercel.ts
danielroe: Great idea. Implemented snapping to the nearest larger size, which feels like the best DX (+ of course warning in dev mode to add desired screen size to config remains).

pi0 (approved):

PR #247, opened 2021-04-30
docs: add info about ctx argument to `getImage` for providers

danielroe:


docs/content/en/5.advanced/1.custom-provider.md
```diff
@@ -28,6 +28,7 @@ export function getImage(src, { modifiers, baseURL } = {}) {
 - `src`: Source path of the image.
 - `modifiers`: List of image modifiers that are defined in the image component or as a preset.
 - `options`: `runtimeOptions` returned by the main provider.
```
pi0, line 30: I think then this is a sub item 😅

PR #256, opened 2021-05-10
fix: remove lazy loading polyfill in favour of native `loading=lazy`

### LINKED ISSUE #190: prefer native lazy loading when possible

### LINKED ISSUE #213: SEO Problem: Images will never be indexed

danielroe:
closes #213, closes #190


pi0:
Additional context: Lazy loading with polyfill support, will be added back with placeholder component (#189)

pi0 (approved):

PR #257, opened 2021-05-11
fix: move ipx middleware to module

danielroe:
* add documentation for manual runtime configuration
* add support for auto config with Nuxt 2.16+ and `.nuxtrc` (depends on https://github.com/nuxt/nuxt.js/pull/9280)


pi0:
One more note: In absence of image module (devDependencies) one needs installing ipx as a dependency! It is making me to think trade off of using normal module and depend on nitro for bundling (there is also fact that we cannot support non-node deployments like CloudFlare workers)

PR #259, opened 2021-05-12
fix: remove leftover `intersectOptions`

danielroe:


pi0 (approved):
Lgtm. We might revert back this when adding lazy support via placeholder component

PR #260, opened 2021-05-12
docs: small tweaks to simplify/clarify docs

danielroe:


docs/content/en/3.api/1.options.md
```diff
@@ -17,13 +17,14 @@ export default {
 
 ## `screens`
 
-List of predefined screen sizes.
```
pi0, line 20: It is still predefined for usage in code

PR #264, opened 2021-05-12
refactor: move `ipx` to runtime servermiddleware

danielroe:
This PR splits out the changes from https://github.com/nuxt/image/pull/257 (without the automatic `nuxtrc` injection, which is pending for now).

* add instructions for runtime ipx
* improve dx with warning
* don't inject ipx if user provides their own ipx mw

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -1,10 +1,8 @@
-import { resolve } from 'upath'
-
+import { relative, resolve } from 'upath'
 import type { Module } from '@nuxt/types'
 import defu from 'defu'
 
 import { setupStaticGeneration } from './generate'
-import { createIPXMiddleware } from './ipx'
 import { resolveProviders, detectProvider } from './provider'
 import type { ModuleOptions, CreateImageOptions } from './types'
 import { pick, pkg } from './utils'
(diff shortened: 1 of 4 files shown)


src/module.ts
```diff
@@ -73,14 +71,37 @@ const imageModule: Module<ModuleOptions> = async function imageModule (moduleOpt
...
 
-  addServerMiddleware({
-    path: '/_ipx',
-    handle: createIPXMiddleware({
-      dir: options.dir,
+  // Only add IPX server middleware if the static/ipx provider is used
+  if (
+    (options.provider === 'static' && nuxt.options.dev) ||
+    options.provider === 'ipx' ||
+    Object.keys(options.providers).includes('ipx')
```
pi0, line 78:
```suggestion
    'ipx' in options.providers
```

src/module.ts
```diff
@@ -73,14 +71,37 @@ const imageModule: Module<ModuleOptions> = async function imageModule (moduleOpt
...
+
+    const hasUserProvidedMiddleware = !!nuxt.options.serverMiddleware.find((mw: { path: string }) => mw.path && mw.path.startsWith('/_ipx'))
+
+    if (!hasUserProvidedMiddleware) {
+      const { createIPX, createIPXMiddleware } = await import('ipx')
+
+      const ipx = createIPX(ipxOptions)
+      addServerMiddleware({
+        path: '/_ipx',
+        handle: createIPXMiddleware(ipx)
```
pi0, line 97: Can we instead generate `.nuxt/image/ipx.js` and register it by path? This makes it working for nitro prod build and making manual instructions easier before 2.16. (for development we can keep doing this until nitro has a better way for shared variable)

PR #281, opened 2021-05-17
docs: document `dir` option

### LINKED ISSUE #237: Document the use of the dir option to define source image location
This is a minor issue but I'm wondering whether its possible to keep images in say `assets/images` rather than `static/images` and be able to use them with with these components?

The reason I ask is everything in `static` gets put in `dist` and deployed, so my original high resolution source images will end up getting uploaded each time and being publicly available (if the url is known). This increases deployment size and time as source images are generally a few megabytes each. Ideally only the derivative generated images would end up in `dist`

tanc:
Closes #237

### DIFF
--- a/docs/content/en/3.api/1.options.md
+++ b/docs/content/en/3.api/1.options.md
@@ -169,4 +169,7 @@ export default {
 }
 ```
 
-Note that changing the `dir` from the default should be done with caution. When using the static provider every image request/variant used during SSR rendering, crawling and generating can be collected and images generated. There are situations where that can't happen: uncrawled pages, modals and tabs, or client-only components. Usually the static provider falls back to the original source url but if it is not available users will get a 404 error.
+**Notes:**
+- For `static` provider, if images weren't crawled during generation (unreachable modals, pages or dynamic runtime size), changing `dir` from `static` causes 404 errors.
+- For `ipx` provider, make sure to deploy customized `dir` as well.
+- For some providers (like vercel), using a directory other than `static/` for assets is not supported since resizing happens at runtime (instead of build/generate time) and source fetched from the `static/` directory (deployment URL)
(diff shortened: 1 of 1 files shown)


docs/content/en/3.api/1.options.md
```diff
@@ -154,3 +154,19 @@ export default {
   }
 }
 ```
+
+## `dir`
+
+Default: `static`
+
+This option allows you to specify the location of the source images if using the `static` provider (IPX).
```
pi0, line 162:
```suggestion
This option allows you to specify the location of the source images when using the `static` or `ipx` provider.
```

docs/content/en/3.api/1.options.md
```diff
@@ -154,3 +154,19 @@ export default {
...
+
+For example you might want the source images in `assets/images` directory rather than the default `static` directory so the source images don't get copied into `dist` and deployed:
+
+```ts [nuxt.config.js]
+export default {
+  dir: 'assets/images'
+}
+```
+
+Note that changing the `dir` from the default should be done with caution. When using the static provider every image request/variant used during SSR rendering, crawling and generating can be collected and images generated. There are situations where that can't happen: uncrawled pages, modals and tabs, or client-only components. Usually the static provider falls back to the original source url but if it is not available users will get a 404 error.
```
pi0, line 172:
```suggestion
**Notes:**
- For `static` provider, if images weren't crawled during generation (unreachable modals, pages or dynamic runtime size), changing `dir` from `static` causes 404 errors.
- For `ipx` provider, make sure to deploy customized `dir` as well.
- For some providers (like vercel), using a directory other than `static/` for assets is not supported since resizing happens at runtime (instead of build/generate time) and source fetched from the `static/` directory (deployment URL)
```

pi0:
Thanks 👍

PR #282, opened 2021-05-18
chore: upgrade docs

Tahul:
Hey 😄

I finally got the `docs/` to work with latest Docus version.

Here is the PR updating documentation structure!

Feel free to comment if anything is missing/needed.

👋


docs/nuxt.config.js
```diff
@@ -1,15 +1,15 @@
...
 
 export default withDocus({
-  docus: {
-    colors: {
-      primary: '#00DC82'
+  rootDir: __dirname,
+  socialImage: {
+    baseUrl: 'DOMIN',
+    chrome: {
+      browserWSEndpoint: 'wss://chrome.browserless.io/'
```
pi0, line 8: Is there any reason why docus is not setting this for default? (PS not sure but `nuxt-extend` should set `rootDir` from top layer if not we can investigate)

pi0: @Tahul

Atinux: Agree this should be set by default

Tahul: Resolved by https://github.com/nuxtlabs/docus/commit/bee93deced36a04f5b3471e6ffa37352400cccaf

docs/package.json
```diff
@@ -1,15 +1,13 @@
 {
   "private": true,
   "scripts": {
-    "generate": "nuxt generate",
+    "generate": "nuxt generate --force-build",
```
pi0, line 4: Is this to detect changes?

Tahul: It was a remainder from Docus docs, it's fixed in last commit. :)

pi0 (approved):
LGTM

PR #299, opened 2021-05-29
feat: add support for netlify provider

danielroe:
This adds support for the Netlify image transformation service.

Note that if you want to use this, you will need to enable [Netlify Large Media](https://docs.netlify.com/large-media/overview/) on Netlify.

More information: https://docs.netlify.com/large-media/transform-images/

### DIFF
--- a/src/runtime/providers/netlify.ts
+++ b/src/runtime/providers/netlify.ts
@@ -1,6 +1,5 @@
 import { joinURL } from 'ufo'
 import type { ProviderGetImage } from 'src'
-import { ImageModifiers } from 'src/types'
 import { createOperationsGenerator } from '~image'
 
 export const operationsGenerator = createOperationsGenerator({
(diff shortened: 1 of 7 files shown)


src/runtime/providers/netlify.ts
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+      modifiers.fit = modifiers.fit || 'contain'
+    }
+  }
+  if (modifiers.format) {
+    // Not currently supported
+    delete modifiers.format
+    if (isDev) {
+      // eslint-disable-next-line
+      console.warn(`Providing format is not currently supported by the Netlify provider. Warning originated from \`${src}\`.`)
```
pi0, line 43: This will cause an unactionable warning for nuxt picture usage.

pi0 (approved):

PR #318, opened 2021-06-11
docs: clarify src origin

bencodezen:
Since developers may be used to managing their images in the `assets` folder, this provides clarity on where to store images and why for developers who may not be as familiar with the build system.


Atinux (approved):
Thanks 💚 

We could maybe link also to https://nuxtjs.org/docs/2.x/directory-structure/static for further details.

bencodezen:
@Atinux Great suggestion. Just pushed the change!

PR #328, opened 2021-06-18
feat: glide provider

### LINKED ISSUE #326: Glide image provider
I am working with a Laravel project using [Glide](https://glide.thephpleague.com/) image optimization to generate urls. I would like to contribute a Glide provider in the next couple days.

shadow81627:
https://glide.thephpleague.com/2.0/api/quick-reference/

resolves #326


src/runtime/providers/glide.ts
```diff
@@ -0,0 +1,32 @@
...
+  joinWith: '&',
+  formatter: (key, val) => encodeQueryItem(key, val)
+})
+
+export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = '/' } = {}) => {
+  const params = operationsGenerator(modifiers)
+
+  if (hasProtocol(src)) {
+    return {
+      url: src
```
pi0, line 25: Why we need this condition?

pi0:
Thanks 🙌 Would you mind also updating playground and docs to add new provider?

src/runtime/providers/glide.ts
shadow81627: Sorry I copied the ipx provider and forgot to remove.

shadow81627:
@pi0 

The only glide instance that I have hosted is the one I use at work... not sure if they would be super happy having it in examples. Do you have any suggestions for the best way to provide a glide instance? Maybe I can host one on Heroku?

pi0:
We usually use examples in provider documentation. For glide i could find this:

https://glide.herokuapp.com/1.0/kayaks.jpg?w=1000&gam=.9&sharp=8

pi0 (approved):
Neat 👍

test/unit/providers.test.ts
```diff
@@ -87,6 +88,17 @@ describe('Providers', () => {
     }
   })
 
+  test('glide', () => {
```
pi0, line 91: /refactor note: We may auto generate tests with a loop (/cc @danielroe)

PR #333, opened 2021-06-23
fix: pass `domains` to runtime options

### LINKED ISSUE #324: IPX runtime not work with external domains
Hi, I was strictly following the instructions found at: https://image.nuxtjs.org/providers/ipx

I checked in many ways, but none of IPX with external domains works.

I created a reproduction: https://codesandbox.io/s/hardcore-euclid-w94ox?file=/pages/index.vue

danielroe:
This no longer treats `static` as unique and moves option normalisation to setup function, which also allows per-provider domains customisation.

closes #324

### DIFF
--- a/src/ipx.ts
+++ b/src/ipx.ts
@@ -5,12 +5,14 @@ import { lt } from 'semver'
 
 import type { ProviderSetup } from './types'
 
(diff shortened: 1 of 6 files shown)


pi0:
Multi-provider purpose is to allow transparently switching between them. For domain is there any special reason to make it vendor-specific?

pi0 (approved):

Issue #338, opened 2021-06-25
Images not generated with `router.base` defined


lapski:
Using  `target: 'static'`, nuxt-img works as expected with `router.base` defined when running `yarn dev`, but images are not generated/working after  `yarn generate`. If I remove `router.base`, it works.

**With** `router.base`:

<img width="415" alt="Screen Shot 2021-06-25 at 12 12 43 AM" src="https://user-images.githubusercontent.com/823871/123368637-2a5d8400-d54a-11eb-932a-2ace4238d097.png">

<img width="205" alt="Screen Shot 2021-06-25 at 12 05 46 AM" src="https://user-images.githubusercontent.com/823871/123368136-3b59c580-d549-11eb-836d-0da19acb95b5.png">

**Without** `router.base`:

<img width="396" alt="Screen Shot 2021-06-25 at 12 13 08 AM" src="https://user-images.githubusercontent.com/823871/123368655-35b0af80-d54a-11eb-9653-9e6dfc60ead8.png">

<img width="320" alt="Screen Shot 2021-06-25 at 12 04 52 AM" src="https://user-images.githubusercontent.com/823871/123368124-35fc7b00-d549-11eb-8f14-5ab2c738fc71.png">

@nuxt/image: "0.4.17"
nuxt: "^2.15.3
Node: v14.16.1

nuxt.config:
```
target: 'static',

buildModules: [
  '@nuxt/image',
],

router: {
  base: '/new'
}

...
```

usage:
```
<nuxt-img
  src="/image.jpg"
  sizes="md:100vw lg:100vw xxl:2000px"
/>
```

Maybe I'm just missing something..?

PR #339, opened 2021-06-25
fix!: prepend router base

btkostner:
This fixes https://github.com/nuxt/image/issues/338 but is a breaking change due to how URLs are handled.

**If you have a default router base of `/`**:
`/images/test.png` points to `~/static/images/test.png` and outputs `/_nuxt/image/hash.png` in the HTML (no difference.)

**If you have a router base of `/app/`**:
`/images/test.png` points to `~/static/images/test.png` and would output `/app/_nuxt/image/hash.png` in the HTML. This is different from the current implementation that would require you to do `/app/images/test.png` to generate correctly.

**If you have a router base of `/app/` with external url**:
`https://example.com/test.png` points to an external URL and outputs `/app/_nuxt/image/hash.png` in the HTML.

## Alternative way to do this

We could remove the changes in `ipx.ts`. Specifying the `src` for images would be the same as it is now, but it would resolve correctly after being built. This would _not_ be a breaking change.


pi0 (approved):
Thanks for PR. LGTM.

src/generate.ts
```diff
@@ -30,7 +30,7 @@ export function setupStaticGeneration (nuxt: any, options: ModuleOptions) {
 
         staticImages[url] = options.staticFilename.replace(/\[(\w+)]/g, (match, key) => params[key] || match)
       }
-      return staticImages[url]
+      return joinURL(nuxt.options.app.basePath, staticImages[url])
```
pi0, line 33: @danielroe Would you please double-check this? Since we are generating into `dist/_nuxt/image` shouldn't it be `cdnURL`?

danielroe: Yes, it should take that into account. Something like:
```suggestion
      return joinURL(nuxt.options.app.cdnURL || nuxt.options.app.basePath, staticImages[url])
```
Plus the following above (with something to replace the resulting `//`?):
```js
publicPath: nuxt.options.app.cdnURL ? '/' : withoutTrailingSlash(nuxt.options.build.publicPath)
```

For reference:
https://github.com/nuxt/nuxt.js/blob/ca020cf9cb4f88a85b06d666c846903d053a8e09/packages/config/src/options.js#L457-L461

btkostner: Updated with suggestion and tests still passing.

btkostner:
Sorry about the force push. I accidentally merged in `master` branch instead of `main`. Should be all fixed now

PR #340, opened 2021-06-25
fix(static): url encoded name included in generated ext

btkostner:
This fixes an issue with the static provider on remote domains including query params in the `nuxt generate` filename outputs.

### Example

Say I have `domains: ['https://images.prismic.io']` in my nuxt config and passing an src like `https://images.prismic.io/blog-system76/30e9de39-20a6-4caf-8728-0ec917942a0f_tumblr_4fef1ab602684355ef9596a0b9fdb122_810bfad7_1280.png?auto=compress,format&rect=56,0,700,700&w=800&h=800`, the generated nuxt image file path would end up looking like `dist/_nuxt/image/181d93.png%3Fauto=compress,format&rect=56,0,700,700&w=800&h=800`.

This PR changes it to output `dist/_nuxt/image/181d93.png` by stripping anything after `%3F` which is the url encoded `?`

### DIFF
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -20,9 +20,10 @@ export function setupStaticGeneration (nuxt: any, options: ModuleOptions) {
     renderContext.image.mapToStatic = <MapToStatic> function ({ url, format }: ResolvedImage) {
       if (!staticImages[url]) {
         const { pathname } = parseURL(url)
+        const [ext] = (extname(pathname) || '').split('%3F')
         const params: any = {
           name: trimExt(basename(pathname)),
-          ext: (format && `.${format}`) || extname(pathname) || '.png',
+          ext: (format && `.${format}`) || ext || '.png',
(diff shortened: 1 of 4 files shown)


btkostner:
An alternative solution for this is to loop over a list of known good formats ('.png', .jpg') and use '.startsWith()' on the ext string.

pi0:
Thanks for PR @btkostner. I've added a refactor to pass original `src` (input) to `mapToStatic` that we can use extracting name and ext before passed to provider. Also added a `guessExt` utility partially implementing your second idea to validate guessed ext against a simple regex.

btkostner:
@pi0 Looks good! I wasn't sure how to add a test for this, but doing `npm link` confirms that this fixes my issue :confetti_ball:

pi0:
Thanks for confarming. Will trigger release after adding few more fixes.

PR #341, opened 2021-06-26
docs: document where to put static images

f3ltron:
I think default behavior is missing for IPX. 

Adding this will add more clarity. But for the alternative serverMiddleware the default behavior is a bit mistyc. Is it recommended to put it inside static folder ?

Because if you do that you have use `/static/myimg` when the default behavior for other is not set the `static`

Don't know what you think about it.

### DIFF
--- a/docs/pages/en/2.components/1.nuxt-img.md
+++ b/docs/pages/en/2.components/1.nuxt-img.md
@@ -25,15 +25,9 @@ Will result in:
 <img src="/nuxt-icon.png" />
 ```
 
-### File Location
-
-Images should be stored in the `static` directory.
-
-For example, the `nuxt-icon.png` referred to in the previous example should be placed in `static` folder under the path `/static/nuxt-icon.png`.
-
-In other words, image stored in the `assets` directory are not compatible with Nuxt Image because those images are managed by webpack.
-
-For more information, you can learn more about the [static directory here](https://nuxtjs.org/docs/2.x/directory-structure/static).
+:::alert{type="info"}
+With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `/static` directrory for making above example working.
+:::
 
 ## Props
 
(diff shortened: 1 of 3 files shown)


pi0:
Thanks @f3ltron for this PR. It totally makes sense. I think we can merge it with previous work (https://github.com/nuxt/image/pull/318) and move both to [default provider sectiction](https://image.nuxtjs.org/getting-started/providers#default-provider)

(because `<nuxt-image>` is documenting component not provider and `ipx` is documenting ipx provider in general not default configuration)

f3ltron:
> Thanks @f3ltron for this PR. It totally makes sense. I think we can merge it with previous work (#318) and move both to [default provider sectiction](https://image.nuxtjs.org/getting-started/providers#default-provider)
> 
> (because `<nuxt-image>` is documenting component not provider and `ipx` is documenting ipx provider in general not default configuration)

I agree but ipx is not checking at static folder by default right? Is it an issue from this provider ?

docs/pages/en/2.components/1.nuxt-img.md
```diff
@@ -25,15 +25,9 @@ Will result in:
...
-
-Images should be stored in the `static` directory.
-
-For example, the `nuxt-icon.png` referred to in the previous example should be placed in `static` folder under the path `/static/nuxt-icon.png`.
-
-In other words, image stored in the `assets` directory are not compatible with Nuxt Image because those images are managed by webpack.
-
-For more information, you can learn more about the [static directory here](https://nuxtjs.org/docs/2.x/directory-structure/static).
+:::alert{type="info"}
+With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `static/` directrory for making above example working.
```
Atinux, line 29: Typo on directory

Atinux:
Except the typo I believe it is a good improvements to the documentation.

We cannot get it perfect at first but through iterations 💚

f3ltron:
Definitely but we are working hard on it :p to make it better. It's already really good

pi0:
(updated the clartification for serverMiddleware  that needs explicitly setting images dir)

Issue #342, opened 2021-06-28
Vercel Provider does not respect domains list


farnabaz:
Vercel image service supports a valid list of domains to load remote images and this domains list generates here https://github.com/nuxt/image/blob/main/src/provider.ts#L24-L38

But the runtime provider of Vercel does not validate remote image domains and pass all images to Vercel images service.

As the result remote images does not load and Vercel images service return 400 Bad Request for remote image.

Vercel provider should validate domains before to detect whether to use Vercel Image provider or use original src, like ipx provider
https://github.com/nuxt/image/blob/main/src/runtime/providers/ipx.ts#L28-L34

PR #343, opened 2021-06-28
fix(vercel): validate remote domains

farnabaz:
Add logic to serve original src if image does not serve from listed domains

Resolves nuxt/image#342

### DIFF
--- a/src/runtime/providers/vercel.ts
+++ b/src/runtime/providers/vercel.ts
@@ -1,12 +1,21 @@
 import { ProviderGetImage } from 'src'
-import { stringifyQuery } from 'ufo'
+import { stringifyQuery, hasProtocol } from 'ufo'
 
 // https://vercel.com/docs/more/adding-your-framework#images
 
-export const getImage: ProviderGetImage = (src, { modifiers, baseURL = '/_vercel/image' } = {}, ctx) => {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for pull-request @farnabaz. I think we shall move this validation out of provider implementation and always validate to externalize URLs when `domains` option is provided and current URL is not matching.

Atinux:
On which file/function can he implanted the validation @pi0

pi0:
I think here we can add the hostname check:

https://github.com/nuxt/image/blob/d9aeae2144ac815f45081d06f1b6b60e62e2ec14/src/runtime/image.ts#L98

farnabaz:
I moved domain validation into `resolveImage` helper and removed logic from both `ipx` and `vercel` providers.

Does it requires other changes @pi0?

pi0:
Changes look good @farnabaz 👌 I'm adding a change to ipx using hostname to use unified method and avoid startsWith check for vercel

pi0:
- Updated `ipx` with hostname check
- Domains are normalized in module
- Only do external check if provider exports `validateDomains` flag enabled for `ipx` and `vercel`. This is to avoid breaking `prismic` and `storyblok` as they accept full URLs (will probably add a flag for this incompatible providers instead later for now flag is undocumented)

PR #380, opened 2021-08-04
feat: move "ipx" into optional dependencies which can be skipped

### LINKED ISSUE #379: Skipping sharp dependency?
Hi there. When using external image optimization providers in a production environment, `sharp` is obsolete, correct? Is there a possibility to skip the installation? The binaries are causing a bit of a headache between various environments.

PatrickHeneise:
Move `ipx` into `optionalDependencies`. These can be skipped with `--no-optional` (npm) or `--ignore-optional` (yarn) when external image optimization APIs are used (in production systems). `sharp` has a few install requirements and binaries which can be tricky in some situations.

Closes #379

### DIFF
--- a/src/ipx.ts
+++ b/src/ipx.ts
@@ -20,12 +20,17 @@ export const ipxSetup: ProviderSetup = async (_providerOptions, moduleOptions, n
     .find((mw: { path: string }) => mw.path && mw.path.startsWith('/_ipx'))
 
   if (!hasUserProvidedIPX) {
-    const { createIPX, createIPXMiddleware } = await import('ipx')
-    const ipx = createIPX(ipxOptions)
-    nuxt.options.serverMiddleware.push({
-      path: '/_ipx',
-      handle: createIPXMiddleware(ipx)
-    })
+    try {
+      const { createIPX, createIPXMiddleware } = await import('ipx')
+      const ipx = createIPX(ipxOptions)
+      nuxt.options.serverMiddleware.push({
+        path: '/_ipx',
+        handle: createIPXMiddleware(ipx)
+      })
+    } catch (err) {
+      console.error('[@nuxt/image] `ipx` is an optional dependency for local image optimization. Please run `yarn install -D ipx` or `npm install ipx --save-dev` or use an image provider. You can find more information here: https://image.nuxtjs.org/providers/ipx.')
+      throw new Error(err)
+    }
   }
 
   // Warn if unhandled /_ipx endpoint only if not using `modules`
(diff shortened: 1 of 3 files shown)


src/ipx.ts
```diff
@@ -20,12 +20,17 @@ export const ipxSetup: ProviderSetup = async (_providerOptions, moduleOptions, n
...
 
   if (!hasUserProvidedIPX) {
-    const { createIPX, createIPXMiddleware } = await import('ipx')
-    const ipx = createIPX(ipxOptions)
-    nuxt.options.serverMiddleware.push({
-      path: '/_ipx',
-      handle: createIPXMiddleware(ipx)
-    })
+    try {
+      const { createIPX, createIPXMiddleware } = await import('ipx')
```
pi0, line 24: Can we please catch error here by chaining import promise? This way we don't show a misleading message if an error happens on `createIPX` call.

PatrickHeneise: `createIPX` will fail either way, no? I'm not 100% sure what you mean with that.

pi0: The second line (L25) is also in try/catch. If an error (other than `ipx` is not installed) when creating ipx instance, it will wrongly show a message that ipx is not installed.

pi0:
Hi @PatrickHeneise sorry for the delay and thanks again for helping on this. It is now merged via #403 and will be released shortly as a fix :) 

BTW some notes:
- Please when making a pull-request, ensure ["Allow edit from maintainers"](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) so that we can directly push cleanup before merge
- It is usually better to avoid doing unrelated changes in a PR (in this case it was recreation of `yarn.lock`)
- If ipx is failing on M1, your Node.js installation is probably old. Which I would recommend [check docs](https://image.nuxtjs.org/getting-started/installation#troubleshooting) to troubleshoot this issue. Even if not using ipx/sharp, it can probably be good to fix root cause on your local machine :)

PatrickHeneise:
Thanks @pi0. I didn't know about the "allow editing for maintainers" option, will make sure that's enabled next time. I'm running Node 16. The problem are cross-arch build issues when Docker is involved.

I'm not familiar with yarn, but when I move/remove something in the dependencies (which happened here), the lock file needs to be updated, as otherwise it's still installed on production systems with `npm ci`. Is that not the case with yarn?

pi0:
Indeed when updating dependencies, yarn lock needs to be updated :)

Issue #276, opened 2021-05-17
secure/signed URL support


pi0:
Some providers like imgix (https://docs.imgix.com/setup/securing-images) or Cloudinary (https://cloudinary.com/documentation/control_access_to_media) support signing URLs to disallow an attacker generating an unlimited amount of URLs causing downtimes, unprivileged access, resource abuse, etc.

To properly supporting this, we need a **server only** mechanism that can sign URLs (if exposing tokens to the client-side, an attacker can still access them to sign!).

nathanchase:
Here's ImageKit's signed URL documentation, for your consideration: 
https://docs.imagekit.io/features/security/signed-urls#generating-signed-urls-on-your-own

shadow81627:
Glide signed URL documentation: https://glide.thephpleague.com/2.0/config/security/

westende:
@pi0 Why the need to wait for Nuxt 3 as you stated in #385? What are the usage limitations you mention?

BenjaminOddou:
@pi0, is it possible to call signed url through Nuxt Image with Imagekit set as provider ?

everyx:
@pi0 @danielroe, has there been any recent progress on this issue?

Baroshem:
@pi0 

I have implemented this feature in Nuxt Cloudinary module and would be happy to implement it for Nuxt Image as well :) 

If you can provide me some guidance from your side how you would like it to work/behave I am happy to create a PR with that feature :)

adamdehaven:
I'd be interested in supporting Cloudflare image signed URLs here as well -- let me know if I can help contribute

PR #385, opened 2021-08-14
feat(provider): imgproxy

### LINKED ISSUE #378: Add support provider "imgproxy"
It's free, fast and opensource stadealone server for resizising images.

https://imgproxy.net/

shadow81627:
resolves #378


shadow81627:
I ran yarn install with the latest version of yarn 1 but it has removed all " from the lock file. I also had to set the test src for imgproxy to a static string of `/test.png` I think there was something changing the signature.

playground/nuxt.config.ts
```diff
@@ -44,6 +44,11 @@ export default <NuxtConfig> {
     imgix: {
       baseURL: 'https://assets.imgix.net'
     },
+    imgproxy: {
+      baseURL: 'https://cc.cz/',
+      key: 'xxxxxxxxxxxxxx',
+      salt: 'xxxxxxxxxxxxxx'
```
ykzts, line 50: Provider options are also included in the pre-built JS that the web browser downloads to generate the image URL on the client side. Therefore, it is possible for a third party to steal the key and salt, but is it safe to leak these values?

shadow81627: > Provider options are also included in the pre-built JS that the web browser downloads to generate the image URL on the client side. Therefore, it is possible for a third party to steal the key and salt, but is it safe to leak these values?

@misaon This PR is implementing https://github.com/nuxt/image/issues/378#issuecomment-898370646

Do you have any insight for the above?

pi0: Related to https://github.com/nuxt/image/issues/276

vonec:
when will this be released :)

casualmatt:
Is this MR still block because there is no way to sign request?

Is possible to see it in the new V1 module, because imgproxy is quite faster then IPX

shadow81627:
I've made my own custom provider for imgproxy to use with Nuxt 3 and Cloudflare. I haven't worried about the imgproxy keys being public, it would be nice to see how to handle the provider being server-side only and how it work at runtime.

Here is the code for my custom nuxt 3 imgproxy provider.

```ts
import { encodeURI } from 'js-base64'
import { Buffer } from 'buffer'
import { OperationGeneratorConfig, OperationMapper } from '~/types/image'
import { joinURL, withBase } from 'ufo'
import hmacSHA256 from 'crypto-js/hmac-sha256.js'
import Base64url from 'crypto-js/enc-base64url.js'
import hex from 'crypto-js/enc-hex.js'

const hexDecode = (hex: string) => Buffer.from(hex, 'hex')

function createMapper(map: Record<string, string>): OperationMapper {
  return (key?: string) => {
    return key ? map[key] ?? key : map.missingValue
  }
}

function createOperationsGenerator({
  formatter,
  keyMap,
  joinWith = '/',
  valueMap,
}: OperationGeneratorConfig = {}) {
  if (!formatter) {
    formatter = (key, value: string) => `${key}=${value}`
  }
  if (keyMap && typeof keyMap !== 'function') {
    keyMap = createMapper(keyMap)
  }
  const map = valueMap ?? {}
  Object.keys(map).forEach((valueKey) => {
    if (typeof map[valueKey] !== 'function') {
      map[valueKey] = createMapper(map[valueKey])
    }
  })

  return (modifiers: { [key: string]: string } = {}) => {
    const operations = Object.entries(modifiers)
      .filter(([_, value]) => typeof value !== 'undefined')
      .map(([key, value]) => {
        const mapper = map[key]
        if (typeof mapper === 'function') {
          value = mapper(modifiers[key])
        }

        key = typeof keyMap === 'function' ? keyMap(key) : key

        return formatter?.(key, value)
      })

    return operations.join(joinWith)
  }
}

const sign = (salt: string, target: string, secret: string) => {
  const msg = hexDecode(salt + Buffer.from(target).toString('hex')) // Uint8Array of arbitrary length
  const hmac = hmacSHA256(hex.parse(msg.toString('hex')), hex.parse(secret))
  const digest = hmac.toString(Base64url)
  return digest
}

const operationsGenerator = createOperationsGenerator({
  keyMap: {
    resize: 'rs',
    size: 's',
    fit: 'rt',
    width: 'w',
    height: 'h',
    dpr: 'dpr',
    enlarge: 'el',
    extend: 'ex',
    gravity: 'g',
    crop: 'c',
    padding: 'pd',
    trim: 't',
    rotate: 'rot',
    quality: 'q',
    maxBytes: 'mb',
    background: 'bg',
    backgroundAlpha: 'bga',
    blur: 'bl',
    sharpen: 'sh',
    watermark: 'wm',
    preset: 'pr',
    cacheBuster: 'cb',
    stripMetadata: 'sm',
    stripColorProfile: 'scp',
    autoRotate: 'ar',
    filename: 'fn',
    format: 'f',
  },
  formatter: (key, value) => `${key}:${value}`,
})

export const getImage: ProviderGetImage = (
  src,
  { modifiers = {}, baseURL = '', key, salt } = {},
) => {
  const encodedUrl = encodeURI(src)
  const path = joinURL('/', operationsGenerator(modifiers), encodedUrl)
  const signature = sign(salt, path, key)

  return {
    url: withBase(joinURL(signature, path), baseURL),
  }
}
```

danielroe:

PR #406, opened 2021-09-08
feat: strapi provider

reliut-g:


src/runtime/providers/strapi.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  if (!breakpoint) {
+    return {
+      url: `${baseURL}/${src}`
+    }
+  }
+
+  return {
+    url: `${baseURL}/${breakpoint}_${src}`
```
shadow81627, line 23: use `withBase` from `ufo`

src/runtime/providers/strapi.ts
```diff
@@ -0,0 +1,27 @@
+import { ProviderGetImage } from 'src'
+
+// https://strapi.io/documentation/developer-docs/latest/development/plugins/upload.html#upload
+
+export const getImage: ProviderGetImage = (src, { modifiers, baseURL = 'http://localhost:1337/uploads' } = {}) => {
+  const breakpoint = modifiers?.breakpoint || ''
```
shadow81627, line 6:
```suggestion
  const breakpoint = modifiers?.breakpoint ?? ''

```

pi0 (approved):

pi0:
Sorry for the late merge.  Implementation seems nice but strapi provider is different from others as it does not actually provide any means to resize or optimize images. Let's have it as is for now!

Issue #409, opened 2021-09-09
[Docs] Homepage light theme text color


HZooly:
It seems that homepage text colors are broken using light theme.

`Optimized images` spam has `text-white` class whereas I'm using light theme.

<img width="1209" alt="image" src="https://user-images.githubusercontent.com/7328625/132770684-31f2382e-094a-4d6e-8ef6-b1c1e5006f29.png">

damisparks:
> Hi, even with #410, light theme doesn't seem to be fixed.
Right @HZooly the light theme has not changed even till now.

the94air:
Same over here, it haven't been fixed since the last time I visited the docs which was weeks ago if not months 😂. It's out of control!!!

the94air:
Pushed a PR #471

marekbrze:
I don't think that it works well right now... 

The text is dark on the dark image. 

![Zrzut ekranu 2022-01-28 o 15 28 46](https://user-images.githubusercontent.com/1033055/151565668-931c8492-616d-4ba0-8151-48becdf942e7.png)

damisparks:
Haha looks like we are messing with the page 😂
Time for another PR.

marekbrze:
Yup, there is a need for an conditional background SVG for light/dark themes;)

marekbrze:
Maybe previously there was just a problem with loading the background SVG?

pi0:
https://github.com/nuxt/image/commit/22d0235baefafc9c23d71dcb9af1ce3a21d41704

Thanks for noticing the issue. We will replace docs soon with a much better package.

Issue #407, opened 2021-09-08
.hero-bg-wrapper is not visible


eliutdev:
.hero-bg-wrapper has a background-image using inline data:image/svg, only works in local

![Screenshot from 2021-09-08 20-38-54](https://user-images.githubusercontent.com/63687573/132566316-e20ad33b-293c-41f5-ad88-736d88ad57dd.png)

PR #410, opened 2021-09-10
chore(docs): update Docus (1.1.x)

Tahul:
Hey 😄, here is a PR fixing the Docus version to 1.1.x.

Please note the Docus API can expect a lot of changes in upcoming weeks/months, so I'll submit this kind of PRs more than once.

If you encounter any problem with that PR, please let me know!

Fixes:
https://github.com/nuxt/image/issues/409
https://github.com/nuxt/image/issues/407

### DIFF
--- a/docs/nuxt.config.js
+++ b/docs/nuxt.config.js
@@ -1,8 +1,32 @@
-import { withDocus } from 'docus'
+import { withDocus } from '@docus/app'
 
 export default withDocus({
   rootDir: __dirname,
-  buildModules: ['vue-plausible'],
+  buildModules: [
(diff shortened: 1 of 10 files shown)


docs/nuxt.config.js
```diff
@@ -1,8 +1,32 @@
...
+    '@docus/twitter',
+    '@docus/github'
+  ],
+  vite: {
+    server: {
+      fs: {
+        strict: false
+      }
+    },
+    optimizeDeps: {
```
pi0, line 17: @Tahul I guess we should do this directly in docus module(s)

PR #416, opened 2021-09-14
feat: emit on image load event for nuxt-img

hansds:
nuxt-image now emits the onLoad event from the contained image

### DIFF
--- a/src/runtime/components/nuxt-img.vue
+++ b/src/runtime/components/nuxt-img.vue
@@ -3,6 +3,7 @@
     :key="nSrc"
     :src="nSrc"
     v-bind="nAttrs"
+    @load="onImageLoad"
   >
 </template>
 
@@ -56,6 +57,11 @@ export default defineComponent({
         // eslint-disable-next-line no-unused-expressions
         this.nSizes
       }
+    }
+  },
+  methods: {
+    onImageLoad (e: Event) {
+      this.$emit('load', e)
     }
   }
 })


shadow81627:
Use `$listeners` to allow binding to all native events. https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components

hansds:
> Use `$listeners` to allow binding to all native events. https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components

Thanks for the suggestion! I updated the PR.

shadow81627:
It would be nice to have a test for the on load event. The docs and playground could also be updated to include the onload event.

src/runtime/components/nuxt-img.vue
```diff
@@ -1,9 +1,5 @@
 <template>
-  <img
-    :key="nSrc"
-    :src="nSrc"
-    v-bind="nAttrs"
-  >
+  <img :key="nSrc" :src="nSrc" v-bind="nAttrs" v-on="listeners" />
```
pi0, line 2: Is it maybe possible directly using `$listeners` here without creating a computed + new object? Also would be nice adding same fix to `nuxt-picture` :)

hansds: @pi0 I plan on doing these changes sometime soon (also the docs and tests @shadow81627 mentioned) - keep you posted

hansds:
Hey @pi0 and @shadow81627 - I fixed your suggestions. Added docs, playground, nuxt-picture and an e2e test.
For the e2e test I wasn't too sure what the best approach was, I opted for verifying a console message, but maybe there's a better way?

Alphability:
@pi0 Do you know when you will release those changes?

madsh93:
@pi0 When will this be released? Docs are live, but `@load` is still not working on latest version.

evangelinepapanicola:
`@load` still doesn't seem to work with nuxt-picture in the latest version. According to the docs this should work - might be worth updating them as it is misleading.

baptistebriel:
Would be great to have @load event on `<nuxt-picture>` 👍🏻

PR #420, opened 2021-09-18
docs(static): ssr need to be true for local image

MrBrown6210:


docs/content/en/1.getting-started/3.static.md
```diff
@@ -3,7 +3,7 @@ title: Static images
 description: 'Optimizing images for static websites'
 ---
 
-If you are building a static site, Nuxt Image will optimize and save your images locally when your site is generated - and deploy them alongside your generated pages.
+If you are building a static site (`ssr` needs to be true), Nuxt Image will optimize and save your images locally when your site is generated - and deploy them alongside your generated pages.
```
pi0, line 6:
```suggestion
If you are building a static site Nuxt Image will optimize and save your images locally when your site is generated - and deploy them alongside your generated pages. (This only works with server-side rendered pages. If you use `ssr: false`, won't work)
```

PR #421, opened 2021-09-19
docs: add info about `animated` modifier

nezzard:


pi0:
Hi @nezzard and thanks for PR. But I guess would be better if we move this example to [ipx provider](https://image.nuxtjs.org/providers/ipx) page or mention in docs. Animated gifs for ipx should work out of the box in the future without this modifier (see https://github.com/unjs/ipx/issues/53 and https://github.com/unjs/ipx/issues/35)

pi0:
With upcoming IPX version, animated for gif is enabled out of the box!

PR #423, opened 2021-09-20
update imgix.md with additional info and links

daletom:
Add an additional documentation section in the imgix providers for Nuxt Image, it links to a recent blog post with an in-depth tutorial on using the Nuxt Image component with imgix.


pi0:
Thanks so much for PR and blog post @daletom 😍

PR #459, opened 2021-12-01
Add high density display support

SimonBackx:
For now, when you specify a size of 320 pixels, it will only generate an image with a size of 320px. But most smartphones have higher density displays, so we also need to add 640px (x2) and 960px (x3) to the srcset attribute (not in the sizes attribute). Adding those sizes to the currently available 'sizes' options, is not a possibility since that will also include it in the sizes attribute (which is not allowed, since that would generate incorrect media conditions)

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -211,7 +211,7 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
 
     for (const factor of highDensityFactors) {
       srcVariants.push({
-        width: _cWidth,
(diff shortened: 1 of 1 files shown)


rwwagner90:
Any updates on this? We really need a way to support 2x and 3x in this package!

jfbloom22:
I was well on my way integrating this plugin, noticed the quality issue on my 4k 24" monitor and am now removing it.  Will be back when high density is supported.

KaragiannidesAgapios:
Any updates on this?

PR #465, opened 2021-12-16
Mismatch between domain and URL on Options

dsclearuc:
Removed the protocol from domain options, since it's not an URL, but domain itself.


pi0:
Thanks! Actually image module automatically strips protocol if set but is much better for consistency in docs now!

PR #471, opened 2022-01-09
Fix hero white text bug

the94air:
Make sure to rename the commit, forgot to make it order by your commit naming


pi0:
Thanks for fix 💚

PR #472, opened 2022-01-09
fix(vercel): sort valid widths before picking largest

harshjv:
Default `.sort` method converts elements into strings and then sorts them in ascending order.

### Before

```
const validWidths = Object.values(sizes || {}).sort()
> [100, 1200, 150]
```

### After

```
const validWidths = Object.values(sizes || {}).sort((a, b) => a - b)
> [100, 150, 1200]
```


pi0 (approved):
Thanks!

PR #477, opened 2022-01-23
feat: support placeholder

### LINKED ISSUE #189: support placeholder

treboryx:
A starting point to resolve #189 

Example added on the playground app.

Suggestions and/or contributions by fellow Nuxters would be appreciated!


src/runtime/components/nuxt-img.vue
```diff
@@ -56,9 +56,23 @@ export default defineComponent({
...
       })
+    },
+    image (): string {
+      return this.placeholder ? this.placeholder : this.nSrc
     }
   },
-  created () {
+  mounted () {
+    if (this.placeholder) {
+      this.image = this.placeholder
```
pi0, line 66: This is a nice start but also means on initial hydration, actual images only start loading when the whole JS bundle is loaded and hydration happens. We can enhance it later with a head script doing the same and hydrating from a `data-` attribute.

treboryx: I suppose hydrating from a `data` atrribute would make a great addition for lazy loading images.

pi0 (approved):

STY1997:
Hi, It seems that the placeholder does not work for cloud providers? I am using Hygraph as the image provider and I get a crashed image icon when loading the image.
Screenshot 2024-05-15 101608
Screenshot 2024-05-15 101702

Issue #483, opened 2022-02-04
External hosted IPX


jeanmatthieud:
Hi,

I want to be able to use an IPX provider, but with an IPX instance hosted on another server than the one running Nuxt.
Currently, the IPX provider source code uses the nuxt `context.base` URL, and cannot be changed :

*/src/runtime/providers/ipx.ts:29*
```js
url: joinURL(nuxtBase, baseURL, params, encodePath(src))
```

I would like to be able to change the base URL (http(s)://xxxx) of the server running the IPX instance.

jeanmatthieud:
Moreover, the `base` property doesn't exists on `nuxtContext`
cf. https://nuxtjs.org/docs/internals-glossary/context/

treboryx:
Hosting your own IPX instance (https://github.com/unjs/ipx) means you're creating a custom provider, therefore you might need to take a look at https://image.nuxtjs.org/advanced/custom-provider

jeanmatthieud:
I believe that after merging ma correction, it will be possible to use the existing IPX provider, but with an external baseURL.

jeanmatthieud:
This is linked to #339

PR #484, opened 2022-02-04
fix(ipx): allow overriding `baseURL` regardless of router base

jeanmatthieud:
#483

Removing this non existing context variable use allows the proper configuration of the `baseURL` variable, and, at the same time, the use of an external IPX server.


pi0 (approved):

pi0:
Thanks. I've fixed types and made modification to still be able completely override `baseURL` still respect router base when generating default `_ipx` endpoint.

chrisspiegl:
I just tried this out and first thought I could not use `baseURL` with the IPX and thus use an external IPX server because the `baseURL` seems to not be included in the typescript type of `IPXOptions`?

Xnapper-2023-02-28-18 54 27

Issue #478, opened 2022-01-29
imgAttrs is not working in nuxt-picture 


vadim-kor:
imgAttrs seem to be ignored when rendering. 
It renders this property to html, but the value is "[Object, object]" instead of string, and it's apliet to the picture tag instead of the imge tag.

ddegooijer:
Have the same issue, any update?

ddegooijer:
Made a PR for the fix, hopefully it will be merged asap.

Khezcute:
Maybe @pi0 can help us pls, I have the same issue

chrisRodgersAirpos:
Experiencing this issue too, has anyone found a fix for it?

PR #491, opened 2022-02-23
fix: imgAttr in nuxt-picture component

ddegooijer:
This is a small bugfix in the nuxt-picture.vue component to enable the imgAttr object to work.  

Link: #478


pi0:
Would you please make a reproduction issue? 🙏🏼

PR #528, opened 2022-06-22
fix(nuxt-img, nuxt-picture): return empty object from `head` when preload is disabled

### LINKED ISSUE #527: Cannot read property 'link' of undefined when not preloading images
Updating from `@nuxt/image` 0.6.0 to 0.6.2 blows up the app with `Cannot read property 'link' of undefined`.

Tracing this back it seems to be because the `head()` method added to support preloading images in `nuxt-img` and `nuxt-picture` doesn't return anything and needs to return an empty object.

gavmck:
Fixes #527 where not preloading images causes the app to throw an error.

### DIFF
--- a/src/runtime/components/nuxt-img.vue
+++ b/src/runtime/components/nuxt-img.vue
@@ -34,6 +34,7 @@ export default defineComponent({
         ]
       }
     }
+    return {}
   },
   computed: {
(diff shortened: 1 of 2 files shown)


pi0 (approved):
LGTM. Thanks.

PR #529, opened 2022-06-22
docs: upgrade docus

### LINKED ISSUE #525: Migrate docs to latest Docus

smarroufin:
Resolves #525


docs/.env.example
```diff
@@ -0,0 +1,3 @@
+# Create one with no scope selected on https://github.com/settings/tokens/new
+# This token is used for fetching the repository releases.
+GITHUB_TOKEN=
```
pi0, line 3: I believe we don't need a token for fetching only content using an unstorage source. It is only needed for listing operations.

pi0:
Thanks for working on this @smarroufin 🔥 I've made few refactor to make it simplified and deployment is working on merging and iterating. Most notably I think we shall add back github and releases with CHANGELOG.md support. Tracking this with @Tahul.

PR #557, opened 2022-07-08
docs: advice `modules` instead of `buildModules`

dosstx:
As of Nuxt 3, `buildModules` property is deprecated. Use `modules` instead.


pi0:
Thanks!

PR #558, opened 2022-07-08
docs: update `static/` dir to `public/`

dosstx:
Different folder structure (static or public) based on Nuxt versions

### DIFF
--- a/docs/content/2.components/1.nuxt-img.md
+++ b/docs/content/2.components/1.nuxt-img.md
@@ -26,7 +26,7 @@ Will result in:
 ```
 
 ::alert{type="info"}
-With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `public/` directory for Nuxt3 and in `static/` for Nuxt2 to make the above example work.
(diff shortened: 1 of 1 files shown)


docs/content/2.components/1.nuxt-img.md
```diff
@@ -26,7 +26,7 @@ Will result in:
 ```
 
 ::alert{type="info"}
-With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `static/` directory to make the above example work.
+With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `public/` directory for Nuxt3 and in `static/` for Nuxt2 to make the above example work.
```
pi0, line 29: Since these docs are targetting Nuxt 3, it is easier to simply mention Nuxt 3 conventions only. Renaming is part of Nuxt 2 => Nuxt 3 migration.

```suggestion
With [default provider](/getting-started/providers#default-provider), you should put `/nuxt-icon.png` inside `public/` directory for Nuxt 3 make the above example work.
```

pi0:
Thanks!

PR #571, opened 2022-07-15
chore: upgrade components to composition API

Diizzayy:
This PR upgrades the `NuxtImg` and `NuxtPicture` components to use the Composition API, and exposes a new `useImage` composable with the same behavior of the `$img` utility.

Other changes:
- the `nuxt-img` component now supports `preload`


AndersNielsen85:
Would be nice adding the example directly to docs. Don't know if that should go to a separate PR? :)

```
<script lang="ts" setup>
const { generate } = useImage()
const bg = computed(() => {
  const imgUrl = generate('/images/colors.jpg', {
    width: 300,
    height: 300,
    quality: 80
  })
  return `url('${imgUrl}')`
})
</script>```

pi0 (approved):
Thanks for helping on this. Nice changes ❤️

pi0:
@AndersNielsen85 I changed back usage to `$img = useImage()` (`$img` is a function with attached props)

Diizzayy:
> Thanks for helping on this. Nice changes ❤️ 

@pi0 no problem, happy to help.

PR #576, opened 2022-07-20
Fixed broken Getting Started link in Docs (v0)

nwittwer:
Currently v0 users are redirected to a 404 page not found page when they click on "Check out the v1.image.nuxtjs.org" link in the [Docs](https://image.nuxtjs.org/getting-started/installation).

This PR fixes the URL to be: https://v1.image.nuxtjs.org/get-started instead of https://v1.image.nuxtjs.org/getting-started/installation.


nwittwer:
Looks like this is fixed already

Issue #608, opened 2022-09-08
Nuxt Picture class prop override Nuxt Image


nthduy:
Context: I wanted to set a different class for picture tag and img tag using `<nuxt-picture>`, based on the [documentation](https://image.nuxtjs.org/components/nuxt-picture#imgattrs) i can do it using imgAttrs prop like so

```html
<nuxt-picture class="my-picture-class" :imgAttrs="{'class': 'my-img-class'}"  src="awsome.jpg"/>
```

Expected result
```html
<picture class="my-picture-class">
   <img src="awsome.jpg" class="my-img-class" />
</picture>
```

Rendered result
```html
<picture class="my-picture-class">
   <img src="awsome.jpg" class="my-picture-class" />
</picture>
```

I think it is related to this [line](https://github.com/nuxt/image/blob/1028381421ddc05e2bd20d0def950ec8216173fc/src/runtime/components/nuxt-picture.ts#L75) in nuxt picture component 

Reproduction:
https://stackblitz.com/edit/nuxt-starter-dtzrru?file=app.vue

PR #620, opened 2022-09-21
fix(nuxt-picture): only pass defined props to `<nuxt-img>` component

nthduy:
Fix https://github.com/nuxt/image/issues/608

Reproduction https://stackblitz.com/edit/nuxt-starter-dtzrru?file=app.vue


hartmut-co-uk (approved):
Looks good to me!
One question comes to mind though - is it correct that `baseImageProps` does not have `class` prop?
And/or is it going to stay like this? 
Maybe it would be safer to explicitly avoid `class` from being passed down?

pi0 (approved):
Thanks ❤️

Issue #584, opened 2022-08-07
Nuxt picture is generating incorrect code.


realandrew:
Edited:
`nuxt-picture` is failing to resize certain formats properly.

Original below:

> I'm using `@nuxt/image-edge v1.0.0-27657146.da85542` with `nuxt v3.0.0-rc.6` (both the latest versions at the time of writing). `nuxt-picture` is generating incorrect `source` elements. Notice (see below) in the `source` element inside `picture` it sets the type to `webp` but is still linking to `jpg` files.
> 
> My code looks like this:
> ```html
> <nuxt-picture
>   :src="String(img)"
>   preload
>   loading="lazy"
>   width="250"
>   height="250"
>   quality="80"
> />
> ```
> 
> Which generates the following:
> ```html
> <picture data-v-d9c38e5e="">
>     <source type="image/webp" srcset="/_ipx/f_webp,q_80,s_320x320/headshot-cropped-1000x1000.jpg 320w, /_ipx/f_webp,q_80,s_640x640/headshot-cropped-1000x1000.jpg 640w, /_ipx/f_webp,q_80,s_768x768/headshot-cropped-1000x1000.jpg 768w, /_ipx/f_webp,q_80,s_1024x1024/headshot-cropped-1000x1000.jpg 1024w, /_ipx/f_webp,q_80,s_1280x1280/headshot-cropped-1000x1000.jpg 1280w, /_ipx/f_webp,q_80,s_1536x1536/headshot-cropped-1000x1000.jpg 1536w, /_ipx/f_webp,q_80,s_1536x1536/headshot-cropped-1000x1000.jpg 1536w" sizes="(max-width: 320px) 320px, (max-width: 640px) 640px, (max-width: 768px) 768px, (max-width: 1024px) 1024px, (max-width: 1280px) 1280px, (max-width: 1536px) 1536px, 1536px">
[...]

PR #625, opened 2022-10-04
fix(picture): Render nuxt picture default img with passed sizes

schnetzi:
🔗 Linked issue
[resolves https://github.com/nuxt/image/issues/584](https://github.com/nuxt/image/issues/584)

❓ Type of change

- [ ] 👌 Enhancement (improving an existing functionality like performance)

📚 Description
This renders the passed sizes of the picture element in the default image source. This will prevent, ensure that the default loaded image has the size of the passed parameters.

📝 Checklist
- [x] I have linked an issue or discussion.
- [ ]  I have updated the documentation accordingly.

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -211,9 +211,11 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
     defaultVar.media = ''
   }
 
+  const defaultSrcWithAdaptedSize = ctx.$img!(input, { ...opts.modifiers, width, height }, opts)
+
   return {
     sizes: variants.map(v => `${v.media ? v.media + ' ' : ''}${v.size}`).join(', '),
     srcset: variants.map(v => `${v.src} ${v.width}w`).join(', '),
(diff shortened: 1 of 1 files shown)


src/runtime/image.ts
```diff
@@ -211,9 +211,11 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
     defaultVar.media = ''
   }
 
+  const defaultSrcWithAdaptedSize = ctx.$img!(input, { ...opts.modifiers, width, height }, opts)
+
   return {
     sizes: variants.map(v => `${v.media ? v.media + ' ' : ''}${v.size}`).join(', '),
     srcset: variants.map(v => `${v.src} ${v.width}w`).join(', '),
-    src: defaultVar?.src
+    src: defaultSrcWithAdaptedSize
```
hartmut-co-uk, line 219: I don't think the default image size should be adapted, also see comment in your original issue https://github.com/nuxt/image/issues/584#issuecomment-1312448414

Issue #618, opened 2022-09-16
[nuxt3] Support responsive image sizes/srcset for `pixel-device-ratio` > 1 (HD and Retina screens, iPad, ..)


hartmut-co-uk:
## Feature

Feature request: Support (nuxt3) nuxt/image with vercel provider for static app -> `nuxt generate`.

### Reproduction example

[https://github.com/hartmut-co-uk/nuxt3-image-pixel-device-ratio-2x-support](https://github.com/hartmut-co-uk/nuxt3-image-pixel-device-ratio-2x-support)

hartmut-co-uk:
As for a solution on how to extend `<nuxt-img>` - how about adding `2x` (`3x?`, ..?) props?

```html
<nuxt-img
            src="/cat.png"
            alt="(5) &lt;nuxt-img&gt; with sizes defined for 1x size & and prop `2x` to also have double pixel density srcset generated"
            class="w-full"
            width="1200"
            height="630"
            2x
            sizes="sm:250px lg:500px"/>
``` 

Alternatives ideas
- prop `density2x` (, `density3` , ...?)
- prop `pixel-device-ratio="1 2 3"`

hartmut-co-uk:
**References** 
- https://github.com/nuxt/image/issues/216
- https://github.com/nuxt/image/pull/459
- https://github.com/nuxt/image/issues/113

hartmut-co-uk:
Would you consider support for this to be added, any preference on which props to go for?

How about adding following boolean props? Or would that make it too restricted? 
Though I think this should solve most use cases?

```ts
  '2x': { type: Boolean, default: undefined },
  '3x': { type: Boolean, default: undefined },
  '4x': { type: Boolean, default: undefined },
```

like for `preload` https://github.com/nuxt/image/blob/db0934c1afb124c579c7c272c0f8c103e360baf5/src/runtime/components/_base.ts#L21

Can I go ahead and create a PR?
https://github.com/nuxt/image/pull/459/files by @SimonBackx seemed mostly complete, but was abandoned?

hartmut-co-uk:
Astro uses `widths` in addition to `sizes` to explicitly define image variants to be provided.
I think this would also be a good + straight option?!?
ref: https://docs.astro.build/en/guides/integrations-guide/image/#widths

SnooHD:
This is a very common issue, and as suggested in the comments above a PR was already created by @simonBackx
Why is this not a priority to fix?

[...]

PR #634, opened 2022-10-19
Add support for densities

marcbelletre:
This PR is based on #459 and enables using a new `densities` attribute to generate larger images for displays with higher resolutions.  

Factors must be separated by spaces and followed by the `x` symbol.   

If the attribute is not specified, the default would be `1x 2x`

This would generate the following sizes : 100px, 300px, 800px:
```vue
<nuxt-img
  src="..."
  sizes="sm:100px md:300px lg:800px"
  densities="1x"
/>
```

This would generate the following sizes : 100px, 200px, 300px, 600px, 800px, 1600px:
```vue
<nuxt-img
  src="..."
  sizes="sm:100px md:300px lg:800px"
  densities="1x 2x"
/>
```

### DIFF
--- a/src/runtime/image.ts
+++ b/src/runtime/image.ts
@@ -2,7 +2,7 @@ import defu from 'defu'
 import { hasProtocol, parseURL, joinURL, withLeadingSlash } from 'ufo'
 import type { ImageOptions, ImageSizesOptions, CreateImageOptions, ResolvedImage, MapToStatic, ImageCTX, $Img } from '../types/image'
 import { imageMeta } from './utils/meta'
-import { parseSize } from './utils'
+import { parseDensities, parseSize } from './utils'
 import { useStaticImageMap } from './utils/static-map'
 
 export function createImage (globalOptions: CreateImageOptions, nuxtContext: any) {
@@ -161,8 +161,10 @@ function getPreset (ctx: ImageCTX, name?: string): ImageOptions {
 function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
   const width = parseSize(opts.modifiers?.width)
   const height = parseSize(opts.modifiers?.height)
+  const densities = opts.densities ? parseDensities(opts.densities) : [1, 2]
   const hwRatio = (width && height) ? height / width : 0
-  const variants = []
+  const sizeVariants = []
+  const srcVariants = []
 
   const sizes: Record<string, string> = {}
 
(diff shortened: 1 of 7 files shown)


hartmut-co-uk:
ref https://github.com/nuxt/image/issues/618

src/runtime/image.ts
```diff
@@ -195,25 +197,45 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
...
+          src: ctx.$img!(input, { ...opts.modifiers, width: _cWidth * density, height: _cHeight ? _cHeight * density : undefined }, opts)
+        })
+      }
+    }
   }
 
-  variants.sort((v1, v2) => v1.screenMaxWidth - v2.screenMaxWidth)
+  sizeVariants.sort((v1, v2) => v1.screenMaxWidth - v2.screenMaxWidth)
+
+  // Remove duplicate size variants,
```
hartmut-co-uk, line 220: how about using `Set` ? 
```ts
const sizeVariantsUnique = [...new Set(sizeVariants)]
```

src/runtime/image.ts
```diff
@@ -161,8 +161,10 @@ function getPreset (ctx: ImageCTX, name?: string): ImageOptions {
 function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
   const width = parseSize(opts.modifiers?.width)
   const height = parseSize(opts.modifiers?.height)
+  const densities = opts.densities ? parseDensities(opts.densities) : [1, 2]
```
hartmut-co-uk, line 164: this would mean a default of `1x 2x` ?

Not sure if 2x should be added by default. This could lead to confusion for many users. 🤔 
Further, it would change the behaviour on upgrade. 

Suggested improvement: allow to configure the default globally via _nuxt.config_.

src/runtime/image.ts
```diff
@@ -195,25 +197,45 @@ function getSizes (ctx: ImageCTX, input: string, opts: ImageSizesOptions) {
...
+  // Loop in reverse order to allow safe deletion
+  for (let i = sizeVariants.length - 1; i >= 0; i--) {
+    const sizeVariant = sizeVariants[i]
+    if (sizeVariant.media === previousSize) {
+      sizeVariants.splice(i, 1)
+    }
+    previousSize = sizeVariant.size
   }
 
+  srcVariants.sort((v1, v2) => v1.width - v2.width)
```
hartmut-co-uk, line 232: Q: shouldn't srcVariants also be de-duplicated?

src/runtime/utils/index.ts
```diff
@@ -91,3 +91,10 @@ export function parseSize (input: string | number | undefined = '') {
     }
   }
 }
+
+export function parseDensities (input: string | undefined = '') {
```
hartmut-co-uk, line 95: should be more strictly typed to return `number[]`

How to deal with invalid input format, maybe it should throw/print some WARN messages and return `[]` if unable to parse...

hartmut-co-uk (changes requested):
Added some notes & suggestions.

Missing:
- docs

Krillko:
I've been waiting for "retina" support since i've first heard about nuxt image.

Can you please please release this sometime

hacknug:
Seems to have been implemented in #769 with tests and a slightly different API. I think it is safe to close this now. //cc @danielroe

danielroe:

PR #661, opened 2022-11-24
fix(nuxt-img): access `prerender.env` only in server side

### LINKED ISSUE #660: [nuxt3] Static image prerendering results in "process is not defined"
Support for prerendering static images was introduced in #614 , but seems to be breaking due to a call to Node's `process`, resulting in a `ReferenceError: process is not defined`.

Minimum reproduction:
https://stackblitz.com/edit/nuxt-starter-4mxwkh

chadsr:
resolves #660

### DIFF
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -86,7 +86,7 @@ export default defineComponent({
       })
     }
 
-    if (typeof process !== 'undefined' && process.env && process.env.prerender) {
+    if (process.server && process.env.prerender) {
       const sources = [
         src.value,
         ...(sizes.value.srcset || '').split(',').map(s => s.split(' ')[0])
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -86,7 +86,7 @@ export default defineComponent({
       })
     }
 
-    if (process.env.prerender) {
+    if (typeof process !== 'undefined' && process.env && process.env.prerender) {
       const sources = [
         src.value,
         ...(sizes.value.srcset || '').split(',').map(s => s.split(' ')[0])


brunodeangelis:
Tested and indeed fixes the issue. Thank you @chadsr

slavanossar:
Could also just do

```js
if (process?.env?.prerender) {
  //...
}
```

chadsr:
> Could also just do
> 
> ```js
> if (process?.env?.prerender) {
>   //...
> }
> ```

Good point... I've switched it over to that syntax.
For some reason I was thinking optional chaining would still throw on `process` being undeclared, but it does not.

uktamka (approved):

Flowko:
how come this not merged yet ?

grad44 (approved):

slavanossar:
> > Could also just do
> > ```js
> > if (process?.env?.prerender) {
> >   //...
> > }
> > ```
> 
> Good point... I've switched it over to that syntax. For some reason I was thinking optional chaining would still throw on `process` being undeclared, but it does not.

Actually sorry, you are right about this. I just ran `nuxt generate` and when loading the page in the browser an error is thrown `ReferenceError: process is not defined`, wasn't happening when checking with `nuxt dev` which threw me off.

chadsr:
> > > Could also just do
> > > ```js
> > > if (process?.env?.prerender) {
> > >   //...
> > > }
> > > ```
> > 
> > 
> > Good point... I've switched it over to that syntax. For some reason I was thinking optional chaining would still throw on `process` being undeclared, but it does not.
> 
> Actually sorry, you are right about this. I just ran `nuxi generate` and when loading the page in the browser an error is thrown `ReferenceError: process is not defined`, wasn't happening when checking with `nuxi dev` which threw me off.

Looks like I feel for the same trap and only ran `nuxi dev` when checking as well.
I've reverted it back to checking with `typeof` now.

src/runtime/components/nuxt-img.ts
```diff
@@ -86,7 +86,7 @@ export default defineComponent({
       })
     }
 
-    if (process.env.prerender) {
+    if (typeof process !== 'undefined' && process.env && process.env.prerender) {
```
pi0, line 89:
```suggestion
    if (process.server && process.env.prerender) {
```

pi0 (approved):
Thanks!

PR #670, opened 2022-12-01
docs: fix typo

madc:
For '[publicPath]/images/[name]-[hash][ext]' it should be:
`/_nuxt/image/file.png` -> `/_nuxt/images/file.png`


pi0:
Thanks!

PR #674, opened 2022-12-03
feat: add static generation support for nuxt picture

jimhlad:
Not sure if this is the right approach, but copied similar generation logic from the nuxt-img component.

It seemed to work first try, so opening a PR in case it is useful.

### DIFF
--- a/src/runtime/components/nuxt-picture.ts
+++ b/src/runtime/components/nuxt-picture.ts
@@ -1,6 +1,6 @@
 import { h, defineComponent, computed } from 'vue'
 import { useBaseImage, baseImageProps } from './_base'
-import { useImage, useHead } from '#imports'
+import { useImage, useHead, useRequestEvent } from '#imports'
 import { getFileExtension } from '#image'
 
 export const pictureProps = {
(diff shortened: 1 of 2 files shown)


FilipBartos:
Changes in this PR fixed image generation in my project. Project is using simple:

```vue
<nuxt-picture
  :src="..."
  :imgAttrs="..."
/>
```

👍

Saquith:
@danielroe Our full static nuxt3 site relies on this PR for release - it affects mobile Lighthouse performance. Is there an estimated release time? Can we assist?

src/runtime/components/nuxt-picture.ts
```diff
@@ -61,6 +62,15 @@ export default defineComponent({
       useHead({ link: [link] })
     }
 
+    if (process.server && process.env.prerender) {
+      const srcKey = nSources.value?.[1] ? 1 : 0
+      const sources = [
+        nSources.value[srcKey].src,
+        ...(nSources.value[srcKey].srcset || '').split(',').map(s => s.trim().split(' ')[0])
+      ].filter(s => s && s.includes('/_ipx/'))
+      appendHeader(useRequestEvent(), 'X-Nitro-Prerender', sources.join(','))
```
pi0, line 71: Refactor: We shall create a utility (`prerenderImages(sources: string[]`) to reuse logic with img/picture components (https://github.com/nuxt/image/blob/200842abdf32aef9d34e39f845b0571103346f19/src/runtime/components/nuxt-img.ts#L89)

src/runtime/providers/ipx.ts
```diff
@@ -12,7 +12,7 @@ const operationsGenerator = createOperationsGenerator({
     quality: 'q',
     background: 'b'
   },
-  joinWith: ',',
+  joinWith: '_',
```
pi0, line 15: Comma support should be supported in upstream nitropack version (today/tomorrow with Nuxt 3.1). More info https://github.com/unjs/nitro/pull/799

mrleblanc101:
@Saquith

How did you use this in your project ? I changed my `package.json` to install from git like so `"@nuxt/image-edge": "jimhlad/image#fix/nuxt-picture-generate",` but I get an error because the built file are not commited and I'm not sure how to build during my own deploy.

pi0:
Thanks for PR! Landed via https://github.com/nuxt/image/pull/725

PR #676, opened 2022-12-06
fix: apply `sizes`/`srcset` when using placeholder

Kolobok12309:
Fix placeholder working with sizes(srcset)

Bug: After mounting, component load only `mainSrc` img, ignoring srcset. For mobiles it load full-size image instead of small version, but after loading, same load small version, total it load 3 images: placeholder, fullsize, smallsize

https://github.com/nuxt/image/blob/bc9ddc0caf4fe32b63f18118ad1b392c7160ddf9/src/runtime/components/nuxt-img.ts#L99-L108

https://github.com/nuxt/image/issues/189#issuecomment-1336961883

I don't understand what this piece is for. Mb update before nextTick?
https://github.com/nuxt/image/blob/bc9ddc0caf4fe32b63f18118ad1b392c7160ddf9/src/runtime/components/nuxt-img.ts#L104


danielroe:
Would you be able to add a regression test for this? 🙏

Kolobok12309:
> Would you be able to add a regression test for this? 🙏

Added simple tests for placeholder with src and with sizes

test/unit/image.test.ts
```diff
@@ -1,10 +1,13 @@
...
 
-import { beforeEach, describe, it, expect } from 'vitest'
+import { beforeEach, describe, it, expect, vi } from 'vitest'
 import { VueWrapper, mount } from '@vue/test-utils'
 
 import { NuxtImg } from '#components'
 
+
+const sleep = (ms = 0) => new Promise(resolve => setTimeout(resolve, ms))
+
```
danielroe, line 10:
```suggestion
```

danielroe (approved):

PR #684, opened 2022-12-17
feat(nuxt-picture): support multiple image formats

### LINKED ISSUE #451: Multiple formats
Avif is better than webp, but some of modern browser doesn't support avif yet, like safari.

Would be great if we could use multiple formats, something like:

```vue
 <nuxt-picture
    src="image.png"
    provider="static"
    :formats="['avif','webp']"
/>


 <nuxt-picture
    src="image.png"
    provider="static"
    :formats="['svg']"
/>
 ```

then returns 

```vue
<picture>
    <source srcset="image.avif" type="image/avif">
    <source srcset="image.webp" type="image/webp">
    <source srcset="image.png" type="image/png">
    <img src="image.png">
</picture>
```

this way, all modern browser will have the best options to display

marcelxpfeifer:
This PR should resolve #451.
I'm open for any kind of feedback or improvement.


marcelxpfeifer:
Just wondering if there is anything that is missing? Any problems that block this feature from being merged? I'm open to any kind of feedback

marcelxpfeifer:
@danielroe any chance we can get this into v1? Anything I can still do to help implement this feature?

danielroe:
odds extremely good

danielroe:
Any chance you could add some tests for this?

marcelxpfeifer:
Yes I will try to do it tomorrow

danielroe (commented):
I might have misread this, but don't we want our `` and the corresponding `<link rel=preload>` to be the non-legacy format?

marcelxpfeifer:
> I might have misread this, but don't we want our `` and the corresponding `<link rel=preload>` to be the non-legacy format?

Oh yes thanks.

src/runtime/components/nuxt-picture.ts
```diff
@@ -94,23 +92,25 @@ export default defineComponent({
...
+          })
+        }),
+        h('img', {
+          ref: imgEl,
+          ..._base.attrs.value,
+          ...(process.server ? { onerror: "this.setAttribute('data-error', 1)" } : {}),
+          ...imgAttrs,
+          src: sources.value[lastSourceIndex.value].src,
+          sizes: sources.value[lastSourceIndex.value].sizes,
+          srcset: sources.value[lastSourceIndex.value].srcset
```
danielroe, line 112:
```suggestion
          src: sources.value[0].src,
          sizes: sources.value[0].sizes,
          srcset: sources.value[0].srcset
```

src/runtime/components/nuxt-picture.ts
```diff
@@ -94,23 +92,25 @@ export default defineComponent({
...
-        ...process.server ? { onerror: 'this.setAttribute(\'data-error\', 1)' } : {},
-        ...imgAttrs,
-        src: sources.value[0].src,
-        sizes: sources.value[0].sizes,
-        srcset: sources.value[0].srcset
-      })
-    ])
+    return () =>
+      h('picture', { key: sources.value[0].src }, [
+        ...sources.value.slice(0, -1).map((source) => {
```
danielroe, line 97:
```suggestion
        ...sources.value.slice(1).map((source) => {
```

src/runtime/components/nuxt-picture.ts
marcelxpfeifer: This would flip it the wrong way. The array is order [...format, legacyFormat]. So avif,webp would return [avif,webp,png]. Right?

danielroe: I think we want the `` rendered to be the non-legacy version. So `[0]` is `avif` in your example. The rest (`webp`, `png`) are the next two, which are rendered above this.

marcelxpfeifer: But the `` is served as the fallback. So it should be the legacyFormat?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_type_attribute

danielroe: but then it seems we were previously (wrongly?) preloading the legacy format image

marcelxpfeifer: I think yes

danielroe:
Thanks! would you take a look at the tests? 🙏

danielroe (approved):

marcelxpfeifer:
I'm so sorry need to fix my local linting and I should work on a different branch till I'm ready. Thanks for your patients and helping out.

danielroe:
Merging but as with some other PRs in this batch, please don't consider the API stabilised until next RC - @pi0 has yet to review and may have some helpful changes to make.

PR #691, opened 2022-12-20
docs: fix cloudflare usage example

baffo:
The example in the documentation for the Cloudflare provider is missing the parent `cloudflare` object. The current example doesn't work on the latest version.


pi0:
Thanks!

Issue #690, opened 2022-12-19
"@nuxt/image-edge": "^1.0.0-27840416.dc1ed65".  still doesn't support static images 


jpengelbrecht:
i find that using.  as Nuxt3 prefers you actually use the assets folder to host your images.
`export default defineNuxtConfig({
  image: {
    dir: "assets/images",
  },
})
`

then using 
` <NuxtImg
          class="desktop"
          src="/website/home/above-fold-illustration.png"
          alt="illustration"
          aria-hidden="true"
        />`
        
doesn't work when deploying static site.   it works on running it on the dev server,  but breaks on the statically built version

"@nuxt/image-edge": "^1.0.0-27840416.dc1ed65",

desenvolvimento02jocc:
Same here.

papadi166:
I'am still waiting for fix xD

aleavikraman:
running into the same issue

Intevel:
Same issue for me!

pi0:
Hi. You should put images to `public/images` directory. assets is a convention for Nuxt to build **build time assets** that get bundled with vite/webpack while image assets are runtime required.

Please let me know if still having issues after moving to public.

ibrahim-990:
i do this but not work with me

export default defineNuxtConfig({
    modules: [
        '@nuxt/image-edge',
    ],
    image: {
        dir: "public/",
    },,    
})

ibrahim-990:
the images shown in mode run nuxt dev but when run nuxt generate the images not shown

ibrahim-990:
and in package.json file 
this
 "dependencies": {
    "@nuxt/image-edge": "^1.0.0-27919678.2f5b64b",
}

Issue #689, opened 2022-12-19
various issues with nuxt image in production


alexiokay:
Error: `File not found (/var/public/images/.../1.jpg)`

`nuxt-img.swiper-image(:src="slide.img" class="swiper-lazy rounded-xl" width="560" height="420" format="webp" alt="logo")`

In development everything works as expected.

danielroe:
would you provide a reproduction please? 🙏

papadi166:
What is reproduction? I provided you the error. That's the code:

` img.swiper-image(:src="slide.img" class="swiper-lazy rounded-xl" width="560" height="420" format="webp" alt="logo")`

danielroe:
Start from this and reproduce the issue: https://stackblitz.com/github/nuxt/starter/tree/v3-stackblitz.

papadi166:
I reproduced it and now it's completly broken... [](https://stackblitz.com/edit/github-hlesub?file=package.json)

Birleyici:
it works locally but not in production (vercel). What's the solution ?

destroytoday:
If it helps, I’m experiencing a similar issue discrepancy when deploying to Vercel, but it disregards any modifiers. (e.g., dev works as expected when specifying width, quality, format, etc., but in production, width is set to an unrelated number (10x the CSS width) and quality is set to 100 (when it's originally set to 70). Fwiw, I’m not specifying any nuxt config `image` settings, in case there’s a missing test case without those settings.

pkgmain:
Having same issue when deploying to Netlify. I notice that my build outputs different routes when the build runs on netlify. When I run `generate` locally I get:
```
ℹ Prerendering 3 initial routes with crawler                                                                                                      nitro 10:55:27
  ├─ / (137ms)                                                                                                                                    nitro 10:55:27
  ├─ /200.html (1ms)                                                                                                                              nitro 10:55:27
  ├─ /404.html (1ms)                                                                                                                              nitro 10:55:27
  ├─ /_ipx/f_webp/avatar.jpeg (45ms)                                                                                                              nitro 10:55:27
  ├─ /_ipx/h_50/go_logo.svg (1ms)                                                                                                                 nitro 10:55:27
  ├─ /_ipx/s_30x50/vue_logo.svg (1ms)                                                                                                             nitro 10:55:27
  ├─ /_payload.js (0ms)                                                                                                                           nitro 10:55:27
✔ You can now deploy .output/public to any static hosting! 
```

[...]

Issue #671, opened 2022-12-02
[nuxt3] Static image generation still results in 404 after latest update for certain images only


jamesxu123:
For some reason images in certain routes are not pre-rendered but not others. This is after #614.

I have found that Nuxt Images with `width` and `height` set do not get rendered statically. E.g. `<nuxt-img format="avif" class="img-fluid" width="1920" height="1000" :src="imgPath" alt=""></nuxt-img>` does not, but `<nuxt-img format="avif" class="img-fluid" :src="imgPath" alt=""></nuxt-img>` does.

jamesxu123:
Example: https://stackblitz.com/edit/github-3bv9kb?file=app.vue

jamesxu123:
Static generation does not seem to be working at all when using Nuxt Picture

jimhlad:
@jamesxu123 Does this only happen when multiple formatters are applied (e.g. width, height, quality, etc)?

If so just made #673 to hopefully solve this issue. And #674 possibly for the second one.

FilipBartos:
@jimhlad I am using simple `<nuxt-picture>` in my project:

```vue      
<nuxt-picture
  :src="..."
  :imgAttrs="..."
/>
```

and your changes in #674 seems to fixed image generation for me.  I hope these changes will be merged soon 👍 

Thanks!

tex0l:
On my end, it doesn't work at all, no static image is built. Am I missing something?

My env:
- `node v18.12.1`
- `macOS 13.0.1 (22A400)`
- `nuxt@3.0.0`
- `@nuxt/image-edge@1.0.0-27827328.bc9ddc0`

I use as well (normally unrelated):
- `@nuxt/content@2.2.2`
- `nuxt-icon@0.1.8`
- `@nuxtjs/tailwindcss@6.1.3` along with `@tailwindcss/line-clamp@0.4.2`
- `sass@1.56.1`

My `@nuxt/image-edge` config is as follows:
```
// nuxt.config.ts
  image: {
    dir: 'assets/images'
  }
```

I use `nuxt-picture` or `nuxt-img` like this:
`<nuxt-img src="image-name.png"/>` or `<nuxt-picture src="image-name.png" />`

And they are rendered as:
```html
<img src="/_ipx/_/image-name.png">
```

 and 

```html
<picture>
  <source
    type="image/webp"
[...]

PR #701, opened 2023-01-03
fix(prerender): Encode URLs to preserve "," in X-Nitro-Prerender Header

posixpascal:
When using IPX provider in combination with `placeholder` attribute the final URL that will get added to the `X-Nitro-Prerender` header contains a plain comma.

Due to the way nitropack currently works this will break static generation of these images due to this [LOC in nitropack](https://github.com/unjs/nitro/blob/main/src/prerender.ts?rgh-link-date=2022-07-06T15%3A08%3A53Z#L260):
```typescript
_links.push(...header.split(",").map((i) => i.trim()));
```

You can reproduce the issue yourself with the latest NPM version of `@nuxt/image-edge` (^1.0.0-27840416.dc1ed65):

```vue
<template>
<div>
     <nuxt-img src="images/logo.svg" /> 
     <nuxt-img src="images/logo.svg" placeholder />
</div>
</template>
```

When using the above vue component only one image will get statically generated when running `yarn run generate`.
The placeholder image gets lost due to the URL being `/_ipx/q_50,s_10x10/images/logo.svg` which nitropack reads as:

- /_ipx/_50
- s_10x10/images/logo.svg

This bug is (closely) related to the following issues:
- #689 
- #690
- #671 

I've also opened a companion PR on nitropack here: https://github.com/unjs/nitro/pull/799

### DIFF
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -90,7 +90,7 @@ export default defineComponent({
       const sources = [
         src.value,
         ...(sizes.value.srcset || '').split(',').map(s => s.split(' ')[0])
-      ].filter(s => s && s.includes('/_ipx/'))
+      ].filter(s => s && s.includes('/_ipx/')).map((source) => source.replace(',', '&'));
       appendHeader(useRequestEvent(), 'X-Nitro-Prerender', sources.join(','))
     }
 
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -90,7 +90,7 @@ export default defineComponent({
       const sources = [
         src.value,
         ...(sizes.value.srcset || '').split(',').map(s => s.split(' ')[0])
-      ].filter(s => s && s.includes('/_ipx/')).map((source) => source.replace(',', '&'));
+      ].filter(s => s && s.includes('/_ipx/')).map(source => source.replace(',', '&'))
       appendHeader(useRequestEvent(), 'X-Nitro-Prerender', sources.join(','))
     }
 


pi0:
Thanks for the PR @posixpascal ❤️ We shall merge this PR after upstream https://github.com/unjs/nitro/pull/799 got merged alongside with Nuxt 3.1 release awaiting Nitropack v2.

pi0:
Hmm thinking more, in the meantime, we might have a faster workaround. IPX supports `&` in addition to `,` for separators (https://github.com/unjs/ipx/blob/ca4d06705ccded4ddfbb453aaf5cb404652e5b0f/src/middleware.ts#L8). Can you please try if we can use an inline replace to only encode `,` as `&` in path segments?

posixpascal:
It does indeed work by just replacing `,` with `&` and without doing the additional decode on nitropack.

I still think a decode on nitropack is necessary when the filename themselves contain `,` but for now replacing `,` with `&` is a good workaround.

I've updated the PR. Let me know if you know a better place to solve the problem as I'm not familiar with the code base yet.

pi0:
Hey. Sorry for delay on this. Since Nitro v2 and Nuxt 3.1 with fix are coming soon (today/tomorrow) let's use your proper upstream fix in https://github.com/unjs/nitro/pull/799.

posixpascal:
I've added the reuqired codestyle changes.

Zielgestalt:
@posixpascal  I can confirm that the images are now generated with '&' as separator in the folders. That's great! Thanks. 
But in the html-files the src is still with a ',' - so the images are not found. Maybe this is another step and not part of this pull request? If so - sorry for disturbing ;-)

pi0:
Thanks for PR! Landed via https://github.com/nuxt/image/pull/725

Issue #682, opened 2022-12-13
@load - Not Firing - Nuxt 3


SleepiestAdam:
Really simple code:

`<nuxt-img format="webp" src="/img/sleepiest-background.png" width="1593" height="717" :class="[backgroundLoaded == true ? 'opacity-100' : 'opacity-0', 'transition-opacity duration-200 ease-in-out']" @load="backgroundLoadComplete" />`

```
<script>
export default {
  data() {
    return {
      backgroundLoaded: false,
    };
  },
  methods: {
    backgroundLoadComplete() {
      console.log('Background did load!');
      this.backgroundLoaded = true;
    },
  },
};
</script>
```

Unfortunately the @load event isn't firing, and I'm not sure why.

Using Nuxt 3 (3.0.0) and @nuxt/image-edge (^1.0.0-27827328.bc9ddc0).

glashtin:
For me it does not call load if the page is called directly or if you hit f5 while on the page.  If the page is called from another page then it works as expected.

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v18.12.1`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@9.2.0`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `app`, `css`, `vite`, `modules`, `colorMode`, `experimental`, `components`, `unocss`, `ssr`, `typescript`, `build`
- Runtime Modules:  `@pinia/nuxt@0.4.6`, `@unocss/nuxt@0.48.0`, `nuxt-icon@0.1.8`, `@nuxt/image-edge@1.0.0-27840416.dc1ed65`, `@vueuse/nuxt@9.9.0`, `@nuxtjs/color-mode@3.2.0`
- Build Modules:    `-`
------------------------------

Der-Alex:
Same here. It seems like the images are cached. Hitting F5 doesn't call the event handler. Hitting ctrl-F5 (delete cache) calls the handler on 'some' images....

piscis:
@pi0 could you have a look at #702 I tried to fix the load listener it should now also work with picture as requested in https://github.com/nuxt/image/pull/416#issuecomment-1132650887

gregorvoinov:
same here, fires only when images are lazy loaded and not inside the viewport on init.

SleepiestAdam:
Any update on a fix being pushed to release? Currently blocking us going into production.

stefbowerman:
[...]

PR #702, opened 2023-01-06
feat: support `load` event for `nuxt-img` and `nuxt-picture`

piscis:
This should implement the Nuxt Image "load" listener for both nuxt image components. It should also solve issue #682


pi0 (approved):
Thanks! LGTM (not tried locally)

PR #709, opened 2023-01-12
refractor: rename layer0 to edgio

rishi-raj-jain:


playground/nuxt.config.ts
```diff
@@ -53,6 +53,7 @@ export default defineNuxtConfig({
       baseURL: 'https://netlify-photo-gallery.netlify.app'
     },
     layer0: {},
```
pi0, line 55: Shall we remove older layer0?

docs/content/5.providers/edgio.md
```diff
@@ -0,0 +1,27 @@
+# Edgio
+
+Optimize images with Edgio's optimization service
```
pi0, line 3: Maybe add something like formerly layer 0 to reduce confusion and improve SEO?

src/types/module.ts
```diff
@@ -51,6 +51,7 @@ export interface ImageProviders {
   imagekit?: any
   imgix?: any
   layer0?: any
```
pi0, line 53: this should be removed too

src/provider.ts
```diff
@@ -21,6 +21,7 @@ const BuiltInProviders = [
   'ipx',
   'netlify',
   'layer0',
```
pi0, line 23: we removed provider...

pi0 (commented):
Thanks. I've added some suggestions mainly to either drop `layer0` from code or make it an alias provider to edgio

playground/nuxt.config.ts
rishi-raj-jain: I think there was no baseURL specified before, shall I update both the configs to have one?

pi0: Good idea

src/provider.ts
rishi-raj-jain: Wouldn't that be breaking one?

pi0: Image v1 is not officially released yet but of course we can also support it as alias if you like...

pi0 (approved):

rishi-raj-jain:
@pi0 Any idea why that docus based redirect is breaking?

PR #730, opened 2023-01-30
fix(cloudinary): update mapping key for `density`

huynl-96:
- update mapping key for cloudinary `density` from `dpi` to `dn`. https://cloudinary.com/documentation/transformation_reference#dn_density
- add support for aspect ratio

image


pi0 (approved):
Thanks!

Issue #759, opened 2023-03-02
Prevent Flicker on Placeholder Replacement


chrisspiegl:
Dear creators / contributors of @Nuxt/Image-Edge,

I am blown away by this project and especially by finding it also finding the self-hostable version of IPX image resizer. Deeply grateful for both.

My main goal was to have the ability for a placeholder image based on base64.

However, I found it rather annoying that there is a short flicker upon replacement of my placeholder.

I tried to figure out where that was coming from, since it was not happening on a previous implementation I was using.

I found that it is related to the `key` param on the image tag. Since it is bound to `key: src` which in turn is a `computed` which is reacting to the `placeholder` which in turn is reacting to the fact of showing/not showing the placeholder.

This has as a consequence that the IMG tag get's re-rendered when the key (aka `src`) changes - which leads to a short flickr while it's being repainted.

But the image is already loaded (which is ensured by the code based `imgEl`.

### Solution

I think the easiest would be to make the `key` instead based on either the `props.src` or `mainSrc`.

Alternatively, I noticed that in the `nuxt-picture` there is no `key` assigned at all? So maybe removing it would be even easier? But I am not sure what the implications would be on that 🙈.

https://github.com/nuxt/image/blob/4cee565c5d0528ac6bc522ea10a06896c1197c5f/src/runtime/components/nuxt-img.ts#L113-L116

I hope this can help make this package even more versatile and the placeholder more beautiful / useful.

vettndr:
Hi @chrisspiegl,
how did you setted up a base64 placeholder?
Are you having issues when the loaded image is switching?

chrisspiegl:
@vettndr, I am storing a Blurhash in the database. And have found a gist script which runs in the browser to turn the Blurhash into a base64 placeholder.

Yes, the issue is when the image is switching but my Pull request above would fix those type of issues.

PR #760, opened 2023-03-02
fix: remove `key` for picture/image to stop flicker when loading

chrisspiegl:
Detailed description / discussion on #759 

Set `key` via `props.src` instead of the computed `src` which leads to redraw of the img element upon replacement of the placeholder with the main image.


danielroe (commented):
This looks good to me - though I'm not sure why there is a key indeed.

I would suggest we remove it from here and `<nuxt-picture>` unless @pi0 thinks it is still required.

danielroe (approved):

pi0 (approved):
LGTM. For context the reason was to support Transitions (https://github.com/nuxt/image/commit/20ed72b376b6845e6e5fd5fa14f698403c43ea34) (@Atinux might remember why we needed and do we still need or not since we were pair programming)

Atinux (approved):
We will see if we need to add it back later but sounds good to me

PR #763, opened 2023-03-04
feat: experimental lcp hints in development mode

### LINKED ISSUE #762: Add Largest Contentful Paint (LCP) hints
Hello there!

What do you think about adding a client only (and dev only) plugin to this module that would utilize the [Performance Observer ](https://web.dev/lcp/#measure-lcp-in-javascript) to locate the potential LCP element and display in the browser console some hints such as:
- If lcp element contains attribute lazy warn user that it will decrease the loading time of LCP, decreasing the performance
- LCP element could have fetchPriority=high attribute to be load as fast as possible
- LCP element could be preloaded in html link
- To achieve better loading time, LCP element could have a better format like webp

This may sound trivial but believe me, I have seen too many production websites with LCP lazy loaded.

Let me know what you think about it @pi0 :)

Baroshem:
Closes #762 

# Largest Contentful Paint hints in development

This PR adds a new client only plugin that will be registered only in development mode. The main purpose of this functionality is to warn users using Image module in the browser console about popular performance issues with images such as:

1. Lazy loading LCP element with `loading="lazy"` attribute
2. Not using modern light image formats such as `webp`
3. Not using `fetchPriority="high"`
4. Not having `width` and `height` attributes set (bad for both LCP and CLS)
5. LCP loading in more than 2500 miliseconds
6. Not preloading the LCP element

## How it works?

The plugin is using [Performance Observer API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver) under the hood to find the potential LCP element and measure metric values such as load time.

Later on, a set of statements is checking whether this LCP element matches the good practices and follows the [Optimize LCP document](https://web.dev/optimize-lcp/) values and recommendations. If not, warnings are displayed in the browser that would allow users to faster find performance bottlenecks and issues. Screenshot below for reference:

Screenshot 2023-03-04 at 11 50 48

## Todo

* Write documentation about this feature
* add configuration option to disable this functionality even in dev environment

## Next steps

Keep in mind that this PR is just beginning of the hints that we can show to the users that would allow them to have a better performing website (and improve Lighthouse and Core Web Vitals score).

I have been thinking about developing a functionality that would fix these kind of issues automatically (something like fontaine plugin) but I failed because there is no way to detect LCP programatically (without access to browser) but maybe someone else here would have some interesting ideas for that.

## Feedback

As always, I am opne for suggestions on what can be improved here. Let me know what can be done better or different way so that we can help users develop more performant websites :)

### DIFF
--- a/docs/content/5.providers/layer0.md
+++ b/docs/content/5.providers/layer0.md
@@ -1,3 +1,0 @@
----
-redirect: /providers/edgio
----
--- a/docs/nuxt.config.ts
+++ b/docs/nuxt.config.ts
@@ -9,4 +9,12 @@ export default defineNuxtConfig({
   plausible: {
     domain: 'image.nuxtjs.org'
   },
+  nitro: {
+    prerender: {
+      ignore: ['/__pinceau_tokens_config.json', '/__pinceau_tokens_schema.json']
+    }
+  },
+  routeRules: {
+    '/providers/layer0': { redirect: '/providers/edgio' }
+  }
 })
--- a/docs/package.json
+++ b/docs/package.json
@@ -8,6 +8,6 @@
   "devDependencies": {
     "@nuxt-themes/docus": "^1.13.0",
     "@nuxtjs/plausible": "latest",
-    "nuxt": "^3.6.1"
+    "nuxt": "^3.6.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test:types": "vue-tsc --noEmit && nuxi typecheck playground && nuxi typecheck example"
   },
   "dependencies": {
-    "@nuxt/kit": "^3.6.1",
+    "@nuxt/kit": "^3.6.2",
     "consola": "^3.2.3",
     "defu": "^6.1.2",
     "h3": "^1.7.1",
@@ -40,7 +40,7 @@
   "devDependencies": {
     "@nuxt/image": "link:.",
     "@nuxt/module-builder": "0.4.0",
-    "@nuxt/test-utils": "^3.6.1",
+    "@nuxt/test-utils": "^3.6.2",
     "@nuxtjs/eslint-config-typescript": "^12.0.0",
     "@types/node": "^18.16.19",
     "@vitest/coverage-v8": "^0.32.4",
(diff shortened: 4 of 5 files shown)


manniL:
A few ideas:

1. Add a config option to enable it, e.g. on staging servers (usually running on PRODUCTION env)
2. Add a hook so people can configure behavior (e.g. forwarding the error) when issues with LCP elements are detected providing the type of issue and ref to the element

harlan-zw:
Nice idea!

Personally, I would say that these types of general performance would be better suited in their own module, or possibly the web vitals module.

If the intent is to have LCP images created correctly, then maybe it's a good idea to introduce a component that handles these things:
- adds the image to be preloaded
- sets the fetch priority
- validates dimensions/format

Couple of thoughts on your code:
- Image format: you're just checking the src, but it's quite possible that they either have no extension or are sending a webp regardless of the extension (i.e Cloudinary does this). Also what if they're using Avif?
- Could be worth checking that the render size of the image matches the actual image dimensions, Lighthouse does this iirc

Baroshem:
Thanks for the feedback @manniL @harlan-zw 

Alex:

1. I have been thinking about adding a `hints` boolean configuration option that when enabled, it will register the plugin even in non development rnvironments. Is this what you meant or would you like it to work a different way?
2. I am not so sure about this new hook. The hints are rather warnings than errors that are giving users hints about what they can do to have better results in LCP and in performance in general. I think this could be an overkill (to implement hooks) for such simple hints only and for now, I cannot find use cases in my head when someone would need to use it in other places in the app (but maybe its just me so if you have any example, please let me know 😀 )

Baroshem:
Thanks @rviscomi !

I think for these more advanced diagnostic metrics we could implement them as next features if this functionality will be useful :)

dargmuesli:
@Baroshem why did you close this PR? :eyes:

Baroshem:
@dargmuesli 

It was there for about 6 months without any update from the package maintainers.

Also, we recently decided to create a package called nuxt hints that would show this and othe rkind of hints to the user so that he can easily make his app better

dargmuesli:
I see, thank you for the explanation!
Related: https://github.com/nuxt/image/pull/989 - happy to hear your feedback if you have any :rocket:

riddla:
> nuxt hints

@Baroshem, where can I find more information about this package?

Baroshem:
Hey @riddla 

This is still in progress and to be honest, I cannot give you any estimation :(

Issue #641, opened 2022-11-08
Nuxt Image sizes attribute doesn't work with Strapi provider


maximelebreton:
Hello!

I'm trying to using **Nuxt Image** with **Strapi** provider but `sizes` doesn't work as excepted.

<details>
  <summary>nuxt.configs.js</summary>
 

  ```javascript
  //...
    image: {
      strapi: {
        baseURL: "http://localhost:1337/uploads",
      },
    }
  ```
</details>

Because when I use this code:
```vue
<nuxt-img
    provider="strapi"
    loading="lazy"
    :src="`${data.file.url}`"
    sizes="sm:100vw md:100vw lg:100vw"
/>
```

the output is:
```html
<img 
    src="http://localhost:1337/uploads/20190131_170142_8085401332.jpg" loading="lazy" 
    sizes="(max-width: 640px) 100vw, (max-width: 768px) 100vw,  (max-width: 1024px) 100vw" 
    srcset="http://localhost:1337/uploads/20190131_170142_8085401332.jpg 640w,
    http://localhost:1337/uploads/20190131_170142_8085401332.jpg 768w,
    http://localhost:1337/uploads/20190131_170142_8085401332.jpg 1024w"
/>
```
As you can see, every `srcset` value **has the same url**...!

And here is the file data coming from **Strapi**:
```json
{
  "url": "/20190131_170142_8085401332.jpg",
  "height": 1440,
  "width": 1440,
  "formats": {
    "thumbnail": {
      "url": "/thumbnail_20190131_170142_8085401332.jpg"
    },
    "medium": {
      "url": "/medium_20190131_170142_8085401332.jpg"
    },
    "small": {
      "url": "/small_20190131_170142_8085401332.jpg"
    },
    "large": {
      "url": "/large_20190131_170142_8085401332.jpg"
    }
  }
}
```

So the **excepted output should be**:
```html
<img 
    src="http://localhost:1337/uploads/20190131_170142_8085401332.jpg" loading="lazy" 
    sizes="(max-width: 640px) 100vw, (max-width: 768px) 100vw, (max-width: 1024px) 100vw" 
    srcset="http://localhost:1337/uploads/small_20190131_170142_8085401332.jpg 640w,
    http://localhost:1337/uploads/medium_20190131_170142_8085401332.jpg 768w,
    http://localhost:1337/uploads/large_20190131_170142_8085401332.jpg 1024w"
/>
```

[...]

PR #765, opened 2023-03-10
strapi local image sharp provider

jiblett1000:
Hello,

I created this originally to solve https://github.com/nuxt/image/issues/641. Strapi doesn't normally provide modifiers beyond "breakpoint". Since the `sizes` prop uses `width` & `height` path/query url parameters to request appropriately sized images, it normally has no effect. However, when the [local image sharp](https://strapi-community.github.io/strapi-plugin-local-image-sharp/guide/modifiers.html#normalize) plugin is installed in Strapi, it provides many modifiers (including width & height). This provides the necessary parameters for the sizes prop to function properly while also allowing additional modifiers through the local image sharp plugin.

This could perhaps be included in the Strapi provider with the functionality being enabled through a nuxt.config.strapi boolean (`localImageSharp: boolean` or similar), but for now, I've separated it into it's own provider. I'm happy to refactor it however after some review and do a PR for the docs as well.

Look forward to hearing from you. Thanks for reading!

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -26,6 +26,7 @@ const BuiltInProviders = [
   'sanity',
   'twicpics',
   'strapi',
+  'strapisharp',
   'storyblok',
   'unsplash',
   'vercel',
--- a/src/runtime/providers/strapisharp.ts
+++ b/src/runtime/providers/strapisharp.ts
@@ -1,0 +1,42 @@
+import { joinURL } from 'ufo'
+import type { ProviderGetImage } from '../../types'
+import { createOperationsGenerator } from '#image'
+
+// https://strapi-community.github.io/strapi-plugin-local-image-sharp/
+
+const operationsGenerator = createOperationsGenerator({
+  keyMap: {
+    width: 'width',
+    height: 'height',
+    resize: 'resize',
+    fit: 'fit',
+    position: 'positon',
+    trim: 'trim',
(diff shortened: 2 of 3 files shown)


pi0:
Thanks for the pull-request. However I'm not sure about advantages of this new preset. Image module already uses default IPX provider which uses local sharp transforms.

jiblett1000:
@pi0 I did not see that before. However, after testing just now, it adds the `_ipx` prefix to images which eliminates the possibility of using it with strapi.

den-kar:
Hi everyone,
@jiblett1000 thx for this PR, looks exactly like the preset I need.
@pi0 the advantage of this new preset is access to modifiers like `format`, `quality`, etc., when using files from the Strapi media library. The current available _strapi_ preset only offers access to the modifier `breakpoint`.

jiblett1000:
@den-kar No problem 😁. In case you haven't seen it. I posted a solution using a custom strapi provider at the bottom of the related issue in case you want to start using it.

Krislunde:
I wish this pull request would get accepted, as it's better than the already existing Strapi provider since it provides utility beyond strapis set "breakpoints" - that way I wouldn't have to roll my own every time. Would there be a higher likelihood of it getting accepted if there was some work done to the existing provider so that an option could provide the functionality this PR includes or something instead of offering two strapi providers?

PR #766, opened 2023-03-10
feat(cloudinary): use `@cloudinary-util/url-loader` to generate urls

### LINKED ISSUE #746: Expanded Cloudinary Features for Nuxt Image
We (Cloudinary) are interested in seeing how we can help support an expanded set of features for the native Nuxt Image component.

The current options when using the Cloudinary provider include some nice critical features, along with a few basic transformations:
- https://github.com/nuxt/image/blob/2f5b64bb5a2ddde2645a5e6fb877d505ebddf34a/src/types/module.ts#L16
- https://github.com/nuxt/image/blob/2f5b64bb5a2ddde2645a5e6fb877d505ebddf34a/src/runtime/providers/cloudinary.ts

But I'd be curious to see how we could implement a wider set of features and what that process may look like for future Cloudinary features.

### Manually Adding Features

We could continue contributing to the mappings and built-in URL formatter. This wouldn't include any dependencies, but would require manually mapping each of these out as needs / requests come along.

### Using a Dependency

I've published a Cloudinary URL Loader with the goal of helping to provide more features in a way that could potentially be compatible with a variety of frameworks, only requiring the dependency update (ideally) for features that work off of non-Nuxt Image props (extended props).

https://github.com/colbyfayock/cloudinary-util/tree/main/packages/url-loader

This could replace some of the existing logic being used in the Cloudinary provider, though the only caveat here is implementing a breaking change or figuring out a way to customize some of the existing features that conflict. Additionally any features that exist in Nuxt Image but not yet the URL Loader could be added.

This however would add a dependency, although small, and I'm not sure if that would be okay from the project point of view and how that would technically work, so that it doesn't just get installed with anyone trying to use Next Image.

I'm also open to suggestions for this library if we end up moving forward with it.

## Example of URL Loader

The URL Loader was built as an abstraction to [Next Cloudinary](https://next-cloudinary.spacejelly.dev/) which was trying to solve a similar goal of adding Cloudinary features on top of the Next.js Image component.

https://github.com/colbyfayock/next-cloudinary/blob/main/next-cloudinary/src/loaders/cloudinary-loader.js#L25

My hope is that we could use this same approach, where I would continue working on the core "props API" from the URL Loader side, and have new, updated, and expanded features propagate through the frameworks that use it.

---

Would love to hear thoughts around this! Ultimately want to see how we can work together moving forward.

I've spoken with Jakub Andrzejewski who's interested in helping out with some of this work, but I'll let him chime in on any thoughts / feedback on the above :)

Baroshem:
Closes #746 

This is a draft PR for now with the idea on how the new cloudinary provider functionality could work. 

This is not a final functionality. I will be updating it in the upcoming days


src/runtime/providers/cloudinary.ts
```diff
@@ -1,111 +1,126 @@
...
   quality: 'auto'
 }
 
 export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = '/' } = {}) => {
+  const cloudName = baseURL.replace('https://res.cloudinary.com/', '').split('/')[0]
   const mergeModifiers = defu(modifiers, defaultModifiers)
-  const operations = operationsGenerator(mergeModifiers as any)
+  // const operations = operationsGenerator(mergeModifiers as any)
 
-  const remoteFolderMapping = baseURL.match(/\/image\/upload\/(.*)/)
```
Baroshem, line 88: @colbyfayock here

Baroshem:
Hey @pi0 

This is a draft PR of the expanded Cloudinary functionalities. For now, I have left commented parts of the code that seems to me as not needed anymore due to the usage of cloudinary url-loader. I asked also Colby about this weird folder mapping if statement and he said that it is done in the url-loader automatically so we do not need it anymore. Same for all operations IMO.

Could you please take a look at it and give me some hints whether this is a good direction? Then, I will make sure to add proper documentation explaing all available options from Cloudinary url-loader package provided by Colby :)

src/runtime/providers/cloudinary.ts
```diff
@@ -1,111 +1,126 @@
...
+//       southEast: 'south_east',
+//       east: 'east',
+//       center: 'center'
+//     }
+//   },
+//   joinWith: ',',
+//   formatter: (key, value) => `${key}_${value}`
+// })
 
 const defaultModifiers = {
```
colbyfayock, line 80: fwiw, i default to format and quality of auto in case you want to additionally clean these up! https://github.com/colbyfayock/cloudinary-util/blob/main/packages/url-loader/src/lib/cloudinary.ts#L141

src/runtime/providers/cloudinary.ts
```diff
@@ -1,111 +1,126 @@
...
   format: 'auto',
   quality: 'auto'
 }
 
 export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = '/' } = {}) => {
+  const cloudName = baseURL.replace('https://res.cloudinary.com/', '').split('/')[0]
```
colbyfayock, line 86: one thought heer, and this could be considered out of scope if it doesnt currently support it, is that someone could be potentially passing in a URL that's not from `res.cloudinary.com` such as if they're using they're own cname

https://cloudinary.com/documentation/advanced_url_delivery_options#private_cdns_and_cnames

i can't remember if my loader technically supports that, i can't remember, but if it does, you could potentially use the parUrl option in the adjacent library (dep of url-loader) @cloudinary-util/util which can grab all the parts, including the host

ill need to add test cases for that to ensure its working properly

src/runtime/providers/cloudinary.ts
```diff
@@ -1,111 +1,126 @@
...
+  //   // If the src is not a remote media file then we need to remove the extension (if it exists)
+  //   src = removePathExtension(src)
+  // }
 
   return {
-    url: joinURL(baseURL, operations, src)
+    url: constructCloudinaryUrl({
+      options: {
+        src: src.substring(1),
+        ...mergeModifiers
```
colbyfayock, line 115: are the width and height getting passed in here? i'm not following if there's a place that they get added to the modifiers as it would, which i wouldn't expect given they're modifiers, but if not it might make sense to pass the width/height through to the options individually?

Baroshem: Yes, they were merged using defu (think of it as merge two object effectively). Now it is not needed as :)

src/runtime/providers/cloudinary.ts
Baroshem: I would say that this is a more advanced scenario that could be added once requested by the community :) At least IMO.

src/runtime/providers/cloudinary.ts
Baroshem: Added :)

Baroshem:
@pi0 Let me know what you think about it :)

docs/content/5.configuration.md
danielroe, line 1: should the first commit in this branch be here?

Baroshem: It shouldn't. Reverted.

danielroe:
This looks nice. My only comments are:

- The docs are not framework-agnostic - the link we point people to is https://next-cloudinary.spacejelly.dev
- This adds an extra ~4.2Mb of install size: https://packagephobia.com/result?p=%40cloudinary-util%2Furl-loader

Baroshem:
@pi0 that sounds good to me @colbyfayock is it ok for you to release it as a part of rc.2?

@danielroe Regarding docs, I think that I can fetch the generated docs from Cloudinary about the url loader into here so that all options will be displayed. But this would require us to keep this part of the docs up to date
regarding bundlesize, what about marking the cloudinary url loader as an optional dependency and intruct the users to download it manually if they want to use cloudinary?

colbyfayock:
im down to follow whateveer makes sense for your workflow, but in terms of timeline, any idea what that would mean?

as far as dependency size goes, honestly this isn't quite my expertise yet, but just showing that the import size is far less and a lot of that comes from additional modules you can import as part of the URL GEn package which i dont

image

Baroshem:
@pi0 what do you think about the timeline for the rc1 and rc2?

@danielroe what do you think about the acreenshot that Colby attached?

Baroshem:
Closing this in favor of https://github.com/nuxt-community/cloudinary-module/pull/111

PR #773, opened 2023-03-18
docs: fix typo

oumarbarry:


Atinux:
The docs for v3 is correct, see https://v1.image.nuxtjs.org/get-started/

We will soon move most of it since I can understand how frustrating it is right now 😬

PR #787, opened 2023-03-29
feat: add `directus` provider

Sandros94:
Lately I started experimenting with [Directus Headless CMS](https://directus.io) and to my surprise it has a nice implementation and usage of [sharp](https://sharp.pixelplumbing.com/).

So I decided to add it as a provider for ease of use.


Sandros94:
> ### ✅ Live Preview ready!
> Name	Edit	Preview	Latest Commit
> **Image**	[Edit on Studio ↗︎](https://nuxt.studio/@nuxt/image/studio/content?refId=%23787)	[View Live Preview](https://v1.image.nuxtjs.org?preview=a4ea070f429839894d7f15b40772c869)	[b5edae4](https://github.com/nuxt/image/commit/b5edae4cb4755e66c2ecdfd7d9fb9bb23aa6da16)

I really can't wait to test nuxt.studio 🥲

danielohling (approved):

chris-lsn (approved):

genu:
LGTM

danielroe:
Would you be able to merge the latest main and add snapshots/expected values for directus? 🙏

Sandros94:
>Would you be able to merge the latest main and add snapshots/expected values for directus?

I've yet never used/created vitest, but there is always a first time right?
I've looked into the commit you did about the tests. Since fully testing directus provider should need a custom test like strapi (with a full directus+db deploy), I should go for the expected values right?

I should add the entry inside the `e2e/__snapshots__/*ssr.test.ts.snap`, add to `provider-coverage.test.ts`:
```ts
const missingProviderTests = [
  'directus', // covered in a unique test
  'strapi', // covered in a unique test
  'layer0' // backwards-compatible alias for edgio
]
```
and what else?

danielroe:
You could add entries for directus in this file:

https://github.com/nuxt/image/blob/28f550d750261c08587d9557a8528661654fb2da/test/providers.ts#L2-L23

Only use the approach I adopted for `strapi` if these entries really don't make sense (strapi doesn't really support most of the options that we were testing there).

If you think it needs _more_ testing, then you can also add additional tests in the `providers.test.ts` file.

Sandros94:
> Only use the approach I adopted for `strapi` if these entries really don't make sense (strapi doesn't really support most of the options that we were testing there).

Directus does support all the listed ones inside `providers.ts`.
But before I add it there, I would like to ask if the `test.png` is needed or not, because Directus uses the file-id from the db to reference to the asset storage, [the actual name of the file isn't needed and its mainly for download and SEO](https://docs.directus.io/reference/files.html#accessing-a-file).

> The location of your actual file originals is based on the project's configuration, but you can consistently access them via the API using the following URL.
> ```
> example.com/assets/<file-id>
> example.com/assets/1ac73658-8b62-4dea-b6da-529fbc9d01a4
> ```
> SEO - You can provide an optional filename after the UUID to optimize for SEO, for example:
> ```
> example.com/assets/<file-id>/<filename>
> example.com/assets/1ac73658-8b62-4dea-b6da-529fbc9d01a4/directus-logo.png
> ```
> This optional filename is also used in the Content-Disposition header when the ?download query parameter is used.

Any options and transformations are added as query parameters to the url.

danielroe:
`test.png` is just meant to be an example file name. It doesn't need to show up in the generated URL if that's not how directus works 👍

Sandros94:
Ok, so the last commit should do it

playground/providers.ts
```diff
@@ -694,5 +694,36 @@ export const providers: Provider[] = [
...
+        alt: '256px width, webp',
+        width: 256,
+        format: 'webp'
+      },
+      {
+        src: 'ad514db1-eb90-4523-8183-46781437e7ee',
+        alt: '256px width, webp',
+        width: 256,
+        format: 'webp',
+        modifiers: { withoutEnlargement: 'true', transforms: '[["blur", 4], ["negate"]]' }
```
danielroe, line 725: these modifiers don't seem to be showing up in the URLs loaded in the e2e test snapshots

danielroe: ahh, I see - you may have added these manually.

in future, you can run `pnpm vitest run -u` to update snapshots

Sandros94: Yes I've added them manually, sorry. First time learning vitest

>in future, you can run `pnpm vitest run -u` to update snapshots

ty

docs/content/5.providers/directus.md
```diff
@@ -0,0 +1,59 @@
...
+<nuxt-img
+  provider="directus"
+  src="ad514db1-eb90-4523-8183-46781437e7ee"
+  :modifiers="{ withoutEnlargement: 'true', transforms: '[[&quot;blur&quot;, 4], [&quot;negate&quot;]]' }"
+/>
+```
+
+::alert{type="info"}
+The `transforms` parameter accepts a two-dimensional JSON array. This means that we should either pass an object to our `modifiers` prop or use `&quot;` to wrap around our parameters in the third quotation layer.
+::
```
danielroe, line 59: would it make sense to pass an array in here and normalise in `getImage` instead?

src/runtime/providers/directus.ts
```diff
@@ -0,0 +1,43 @@
...
+    withoutEnlargement: {
+      true: 'true',
+      false: 'false'
+    }
+  },
+  joinWith: '&',
+  formatter: (key, value) => `${key}=${value}`
+})
+
+export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL = 'http://localhost:8055/assets' } = {}) => {
```
danielroe, line 38: I think it makes sense for a user to have to provide this baseURL in the config - let's not hard-code it in the provider.

Sandros94: True, I did it when I was still mainly using [intevel/nuxt-directus](https://github.com/intevel/nuxt-directus)

danielroe: Understood. I think we should update that one as well for the same reason.

docs/content/5.providers/directus.md
Sandros94: Probably yes, this was one of those things that I didn't much tuch because of lack of experience in js/ts. Leaving it more barebones for the *experts* to decide.

src/runtime/providers/directus.ts
Sandros94: What about:

```ts
export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL } = {}) => {
  if (baseURL && !baseURL.endsWith('/assets')) {
    baseURL += '/assets'
  }
  return {
    url: joinURL(baseURL || 'http://localhost:8055/assets', src + (operations ? ('?' + operations) : ''))
  }
}
```

This way we both handle defined baseURLs (even if the developer forgets to add `/assets`) but still fallback to a default one?

docs/content/5.providers/directus.md
danielroe: you're doing a great job!

Sandros94: later I'll push couple of commits to potentially solve this

Sandros94: I'm having an hard time, but I'm starting to think that I forgot how to do the request to Directus correctly, because in a dev test I spun up if I do `http://localhost:8055/assets/ad514db1-eb90-4523-8183-46781437e7ee?transforms=%5B"blur"%2C"4"%5D` I get:

```json
{
    "errors": [
        {
            "message": "\"transforms\" Parameter does not allow \"b\" as a transformation.",
            "extensions": {
                "code": "INVALID_QUERY"
            }
        }
    ]
}
```

and I don't understand why I get `\"b\" instead of blur...

Sandros94: Ok I've figured out how to do it, and it has been quite a nice challenge to face. But I'm happy that it fully works.

I'm resolving this with [51edbd7](https://github.com/nuxt/image/pull/787/commits/51edbd77a72e44a0334d7fa1c56b49e4c7cc446d)

Sandros94:
I wanted to test things one last time in the playground, but I cannot make it to work anymore (last time was when I created the PR).

I do have corepack enabled and I'm following the readme, I see that a node process start to work but the nuxt server never actaully starts 🫤

It hangs here:
image

danielroe:
Would you run pnpm i to ensure you're running the latest dependencies?

Sandros94:
>Would you run pnpm i to ensure you're running the latest dependencies?

Pulled latest commits, `pnpm i` and now is working.

I started a Directus test via docker and the requests are perfectly fulfilled.
IMO we are ready to go

danielroe (approved):

docs/content/5.providers/directus.md
```diff
@@ -0,0 +1,62 @@
...
+<nuxt-img
+  provider="directus"
+  src="ad514db1-eb90-4523-8183-46781437e7ee"
+  :modifiers="{ height: '512', withoutEnlargement: 'true', transforms: ['blur:4', 'negate'] }"
+/>
+```
+
+::alert{type="info"}
+Directus exposes a `transforms` parameter that accepts a two-dimensional JSON array to communicate directly with the [sharp API](https://sharp.pixelplumbing.com/api-operation). However, for ease of use, we instead decided to expose a single-dimensional array, passing a list of strings as `key:value` or just `key`.
+::
```
rijkvanzanten, line 62: Heya! Directus maintainer here 👋🏻 

I'm wondering if this should either mirror the Directus type signature exactly to be consistent (so it's easier to learn/ask/find how to use this) _or_ go fully the other way and make this super Vue / JS friendly by supporting it as an object type (like `{ blur: 4 }` instead of `['blur:4']`). I feel like the current solution is a bit of both, but not quite enough of either 😄

danielroe: That sounds good to me. I'd recommend we mirror directus so the docs take less 'translation'

Sandros94: Currently I'm a bit split right now (as someone that just got into js/ts):
On one hand, quickly managing a simple list of _secondary_ transforms as a set-and-forget thing is quite appealing. As well as it would be managing a full object for custom scripting.
On the other hand, while the list of lists was easy as an object to deal with for scripting, I've found it quite challenging at first that I had to `[["key",value],["key"]` to do things around. Even more once I got back to this PR and I had forgotten that I was dealing with a list of lists and not with an object.

This is especially true when I'm not using `nuxt-image`. Where maybe I would like to hardcode a transform like `%5B%5B%22blur%22%2C4%5D%2C%5B%22negate%22%5D%5D` into a component to bake a simple effect.

I'm surely biased to go for the simple list or the object, but I do see what could be the "ah ok" confusion between the two docs. ngl I though about going into directus and sharp repos to find out _why_ a list of lists. But currently my js/ts muscles are still too weak. 🙃

Sandros94: I would say that this should depend on what type of developer we would like to target.

- **the junior:** the one that likes simple lists.
- **the flex:** the object fan.
- **the pro:** the docs enjoyer.

rijkvanzanten: > On the other hand, while the list of lists was easy as an object to deal with for scripting, I've found it quite challenging at first that I had to [["key",value],["key"] to do things around. Even more once I got back to this PR and I had forgotten that I was dealing with a list of lists and not with an object.

The reasoning behind that "double" array is that we don't know nor "care" how many arguments a given function has, so the format is `[function, ...args]`. That in turn means we can expose everything sharp has to offer as a catch-all fallback for the couple of functions we don't make available as the "simple" / friendly way (yet). 

---

Love the analogy, but I don't think it's totally accurate in this case 😄 To me, the "simple" parameters that already exist (eg width/height/format/etc) cover the "junior" and "flex" use case, whereas the transforms should really be seen as a catch-all last-resort for more powerful access to sharp directly. When it comes to that last use case, I'm worried that adding an additional abstraction that only exists in the context of nuxt/image is going to do more harm than good as now it's neither a "nuxt" nor a "directus" thing (which makes it hard to discover/learn) 🙂

Sandros94: > The reasoning behind that "double" array is that we don't know nor "care" how many arguments a given function has, so the format is `[function, ...args]`.

True! So that would mean that if I want to use the `normalise` function (that needs two values, a lower and an upper limit) it would become `['normalise', 20,75]` right?

If this is the case I might prefer to revert the array normalization changes and try to simplify the examples in the nuxt-image/docs. @danielroe what do you think?

Sandros94: @rijkvanzanten look at this changes [4143b65](https://github.com/nuxt/image/pull/787/commits/4143b652573751eef6661414e8cb49fcc20dc541):

Basically I'm directly managing the transform as a list of lists and encoding it in urlsafe automagically.


p.s.: I'm noticing that I don't know how to reference line changes in a comment or suggest a review for a specific commit.

danielroe: > If this is the case I might prefer to revert the array normalization changes and try to simplify the examples in the nuxt-image/docs.

Perfect 👍

docs/content/5.providers/directus.md
```diff
@@ -58,5 +58,5 @@ Since Directus exposes [the full sharp API](https://sharp.pixelplumbing.com/api-
 ```
 
 ::alert{type="info"}
-Directus exposes a `transforms` parameter that accepts a two-dimensional JSON array to communicate directly with the [sharp API](https://sharp.pixelplumbing.com/api-operation). You can easily manage any transforms as a list of lists.
+Note that the `transform` parameter, as stated in the [Directus documentation](https://docs.directus.io/reference/files.html#advanced-transformations), is basically a list of lists. This is to facilitate the use of those sharp functions, like [`normalise`](https://sharp.pixelplumbing.com/api-operation#normalise), that would need multiple values in input: `transforms: [[normalise, 1, 99], ['blur', 4], ['negate']]`.
```
danielroe, line 61:
```suggestion
Note that the `transform` parameter, as stated in the [Directus documentation](https://docs.directus.io/reference/files.html#advanced-transformations), is basically a list of lists. This is to facilitate the use of those sharp functions, like [`normalise`](https://sharp.pixelplumbing.com/api-operation#normalise), that would need multiple values in input: `transforms: [['normalise', 1, 99], ['blur', 4], ['negate']]`.
```

docs/content/5.providers/directus.md
Sandros94: Ok, marking this one as resolved too. We are back to a list of lists and made sure to encode it properly.

Sandros94: > Hehe getting a bit off-topic in this PR, but do make sure to upvote and leave your thoughts on https://github.com/directus/directus/discussions/13041 🙂

This PR made me think of a nice usecase, I'll make sure to write it in that discussion (that I was still reading thoroughly).

> You can (shift-)click the line-numbers in the file to update the URL in the files tab to be a permalink to those lines 🙂
>
> https://github.com/nuxt/image/pull/787/files#diff-5ce8ceb4b0a441713600d54be499a9d005b00077aa2ef59009d492eb6fa422d7R43-R44

See? It graphically didn't render, but maybe it's a github thing. I'll investigate in a different moment.

docs/content/5.providers/directus.md
```diff
@@ -0,0 +1,62 @@
...
+```vue
+<nuxt-img
+  provider="directus"
+  src="ad514db1-eb90-4523-8183-46781437e7ee"
+  :modifiers="{ height: '512', withoutEnlargement: 'true', transforms: [['blur', 4], ['negate']] }"
+/>
+```
+
+::alert{type="info"}
+Note that the `transform` parameter, as stated in the [Directus documentation](https://docs.directus.io/reference/files.html#advanced-transformations), is basically a list of lists. This is to facilitate the use of those sharp functions, like [`normalise`](https://sharp.pixelplumbing.com/api-operation#normalise), that would need multiple values in input: `transforms: [['normalise', 1, 99], ['blur', 4], ['negate']]`.
```
danielroe, line 61:
```suggestion
Note that the `transforms` parameter, as stated in the [Directus documentation](https://docs.directus.io/reference/files.html#advanced-transformations), is basically a list of lists. This is to facilitate the use of those sharp functions, like [`normalise`](https://sharp.pixelplumbing.com/api-operation#normalise), that would need multiple values in input: `transforms: [['normalise', 1, 99], ['blur', 4], ['negate']]`.
```

PR #826, opened 2023-05-22
Properly hash images based on contents and allow skipping generation of images

jthawme:
This is for nuxt 2 and recognise it may too much trouble to have merged in, but i wanted to open the PR in case it was useful for the core package, or the wider community

I patched the code so that it actually hashes the image based on the contents, not the filename, as that is more useful.

And on top of that, I've also added a method to the options to allow for skipping generation for files, couple with the contents hashing mentioned above, this becomes a very powerful method to speeding up build times potentially

The npm package of the fork is here, if anyone needs to use this functionality in their application
https://www.npmjs.com/package/jthawme-nuxt-image


danielroe:

PR #830, opened 2023-05-26
test: add back and update tests

danielroe:
This can undoubtedly be improved but the aim is to add back the previous test suite and provide a scaffold for regression testing as we merge and iterate before the release of v1 of `@nuxt/image`.


danielroe:
Merging to iterate - feedback and improvements welcome @pi0 ❤️

PR #836, opened 2023-05-29
perf: do not inject `$img` by default

danielroe:


pi0 (approved):
LGTM. I have added a (dev only) placeholder to instruct users about migration since this is a fairly common feature and many are probably using it...

PR #840, opened 2023-05-31
feat: add `none` provider and only enable `ipx` when node present

danielroe:


src/provider.ts
```diff
@@ -108,9 +110,7 @@ export function detectProvider (userInput?: string) {
     return userInput
   }
 
-  if (process.env.VERCEL || process.env.VERCEL_ENV || process.env.NOW_BUILDER) {
+  if (provider === 'vercel' || process.env.VERCEL_ENV) {
```
pi0, line 113: Since now we are migrating to `std-env`, i would completely depend on it to reduce fractions. (upstream: https://github.com/unjs/std-env/issues/54)

```suggestion
  if (provider === 'vercel') {
```

danielroe:
Totally happy to rename before release and not keep alias - feel free to push change directly if you like :+1:

PR #848, opened 2023-06-05
fix(pkg): use esm build as `main` field

oraphadev:
This PR should resolve https://github.com/nuxt/image/issues/379
I'm open for any kind of feedback or improvement.


pi0 (approved):
LGTM as quick fix thanks.

We shall add `exports` field and once added, revert back main to cjs because all modern node runtimes will pick `exports` (also in CJS environment)

PR #870, opened 2023-06-14
fix(storyblok): don't apply resize/format/filter on vector

RyanMulready:
Given the nature of headless CMS we sometimes don't know what image type will be passed into `nuxt-img`. This prevents any resizing, formatting and filtering on vector images. Today if any of these are applied storyblok either responds with a `40X` error or the image appears distorted.


danielroe (approved):

danielroe:
Would you lint the commit? 🙏

Issue #657, opened 2022-11-24
[nuxt3] Feature Request: config option for default `format` & `quality`


hartmut-co-uk:
I'd like to configure a default image _format_ for all images via nuxt config:

```ts
export default defineNuxtConfig({
    modules: [
        '@nuxt/image-edge'
    ],
    image: {
        // 
        format: "webp",
        quality: "80"
    },
}
```

#### Questions
1. According to configuration, this might be possible via provider specific default modifiers?
1. Is this the best choice for such basic props as 'format', 'quality'?
1. Is it also possible (to add) for IPX? ([IPXOptions](https://github.com/unjs/ipx/blob/c8094bce35bee12e04ad167fee983cd7789c97fa/src/ipx.ts#L30))

#### Refs
- https://v1.image.nuxtjs.org/components/nuxt-img/#format
- https://v1.image.nuxtjs.org/configuration/#modifiers
- https://github.com/unjs/ipx/blob/c8094bce35bee12e04ad167fee983cd7789c97fa/src/ipx.ts#L30

hartmut-co-uk:
I've tested with following config:

```ts
export default defineNuxtConfig({
    modules: [
        '@nuxt/image-edge'
    ],
    image: {
    provider: 'ipx',
    ipx: {
      modifiers: {
        quality: '80',
        format: 'webp',
      }
    }
  }
}
```

This works just fine and is applied to all nuxt-img as a default, also can be overwritten if required.

=> So I think it's actually just a missing type def in IPXOptions?

sungeunl:
is it working fine ?

hartmut-co-uk:
at the time of writing it was working, I no longer can confirm the exact version though...

madebyfabian:
I can confirm, still works with `"@nuxt/image": "1.0.0-rc.3"`. But still with a type error.

Baroshem:
Both configuration options are now available https://image.nuxt.com/get-started/configuration#quality

PR #880, opened 2023-06-19
feat: allow setting global `format` defaults in config

Tragio:
Hi there 👋 

At the moment, we can only define formats for Nuxt Picture at the component level. This PR would also allow configuring the Nuxt Picture at the global level.

I'm only now getting into Nuxt Module development (kudos to @danielroe streams), so let me know if is there anything that can be done in another way or in a better way. 💪 

Thank you very much for the amazing work and community 🙏


src/runtime/components/nuxt-picture.ts
```diff
@@ -29,8 +29,8 @@ export default defineComponent({
 
     type Source = { srcset?: string, src?: string, type?: string, sizes?: string }
     const sources = computed<Source[]>(() => {
-      const format = props.format || (originalFormat.value === 'svg' ? 'svg' : 'webp')
-      const formats = format.split(',')
+      const format = props.format || $img.options.format?.[0] || (originalFormat.value === 'svg' ? 'svg' : 'webp')
```
danielroe, line 32:
```suggestion
      const format = props.format || (originalFormat.value === 'svg' ? 'svg' : ($img.options.format?.[0] || 'webp'))
```

hartmut-co-uk:
It would be great if we also could add `quality` and apply it to `<nuxt-image>`.
ref https://github.com/nuxt/image/issues/657

For nuxt-image we'd use the first elem `format[0]`..?
(This could be a separate PR)

src/runtime/components/nuxt-picture.ts
Tragio: Ahh yes yes, of course. Thank you! 😄

danielroe (approved):

danielroe:
@hartmut-co-uk That makes sense to me too.

danielroe:
Yes, it's a separate PR 👍

hartmut-co-uk:
Could you also add unit tests for the new config?

Test cases brainstormed: 
- with global config, no prop -> global config is used
  - global config, single format -> all expected image variants
  - global config, multiple formats -> all expected image variants in all formats
- with global config, with prop -> prop is used (takes precedence over global)
- global config invalid value -> Error thrown (??) || or warning logged & ignored?!?
- no global config, no prop -> default applies ('webp'?)
- global config is ignored for 'svg'

@danielroe do we have an example of setting module config to be used in unit tests...?

danielroe (approved):

danielroe:
Thank you both ❤️

PR #884, opened 2023-06-21
feat: allow setting global `quality` defaults in config

Tragio:
Hi there 👋 

This PR allows setting the quality of the generated images in the module configuration.

Thank you very much for the amazing work 🚀

### DIFF
--- a/docs/content/5.configuration.md
+++ b/docs/content/5.configuration.md
@@ -31,6 +31,20 @@ export default defineNuxtConfig({
 })
 ```
 
+## `quality`
+
+The quality for the generated image(s).
+
+You can also override this option at the component level by using the [`quality` prop.](/components/nuxt-img#quality)
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  image: {
+    quality: 80,
+  }
+})
+```
+
 ## `format`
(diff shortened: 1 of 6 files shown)


danielroe:
You are a star - thank you ❤️

danielroe (approved):

hartmut-co-uk (changes requested):
Great!
The solution should also apply `quality` at module config level to `<nuxt-picture>`, right?

Could you please also add tests to `test/unit/picture.test.ts`?

Tragio:
> Great! The solution should also apply `quality` at module config level to `<nuxt-picture>`, right?
> 
> Could you please also add tests to `test/unit/picture.test.ts`?

Thank you very much 🙏  I only added one test, should I add all the tests that I did for `nuxt-img`? Just don't want to be redundant, since all is being [passed to `nuxt-picture`](https://github.com/nuxt/image/blob/main/src/runtime/components/nuxt-picture.ts#L19). If you want I can do it without any problem 🙂

hartmut-co-uk:
Fair enough, I think this is good to go! 💪

hartmut-co-uk (approved):

PR #886, opened 2023-06-21
nuxt-picture: allow parsing format as array

marcelxpfeifer:
To better align the image format prop with the global setting for format, I now allow parsing an array as prop as well. Also minor tweaks to the docs and tests.


hartmut-co-uk (changes requested):
This looks good to me overall. 
Extra ⭐ for the docs! :-)

Could you address the failing type checks? (`pnpm test:types`)

src/runtime/components/_base.ts
```diff
@@ -109,3 +118,47 @@ export const useBaseImage = (props: ExtractPropTypes<typeof baseImageProps>) =>
     modifiers
   }
 }
+
+export const useBasePicture = (props: ExtractPropTypes<typeof basePictureProps>) => {
```
danielroe, line 122: For simplicity and reduced bundle size, do you think we can avoid splitting out `useBasePicture` and instead override the props directly in `nuxt-img`?

PR #769, opened 2023-03-14
feat: support `densities` for devices with `DevicePixelRatio` > 1

Mihanik71:
Add densities param for issues #216 

Add:

- default value for densities in nuxt.config
- add param for img and picture components

resolves #216
closes https://github.com/nuxt/image/pull/803

PR #928, opened 2023-07-24
fix: respect densities prop with `<nuxt-picture>`

### LINKED ISSUE #927: nuxt-picture don't support densities setting, only on nuxt-img
I want to force the densities to 1x on nuxt-picture. Unfortunately, it can only be set on nuxt-img. However, we need to use both webp and jpeg formats simultaneously, so we have to use nuxt-image.

ausir0726:
resolves #927 

and I saw the merged commit https://github.com/nuxt/image/pull/769/commits/bc3d333a7149e4fc62007685072300d5cba33cd1#diff-3fc0f6b5f15d4b824465eac0cc09094d1a5889ad6867b3b9e908512b5aab47ddR47
 not include nuxt-picture compoment.


ausir0726:
Is there anything I can do to keep this going?

danielroe (approved):

PR #955, opened 2023-08-29
feat: add `apiVersion` to cloudimage provider

TiborPL:
Newly created Cloudimage tokens don't need `v7` in url which was hardcoded.
Made optional option `apiVersion`. ~~Kept `v7` as default so not to break old codes.~~

Also clarified the documentation regarding new tokens and api version.

## 👉 Migration

If you have been using `cloudimage` provider with a legacy token, you will need to add the `apiVersion: 'v7'` to your nuxt.config:

```diff
  export default defineNuxtConfig({
    modules: ['@nuxt/image'],
    image: {
      cloudimage: {
        token: 'demo',
        baseURL: 'sample.li',
+       apiVersion: 'v7'
      },
    }
  })
```


docs/content/3.providers/cloudimage.md
```diff
@@ -47,10 +45,31 @@ export default defineNuxtConfig({
...
+
+**Note:** For tokens created after **20.10.2021**, `apiVersion` need to be set to empty string or you can use `cdnURL` with value `https://{token}.cloudimg.io`.
+
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  image: {
+    cloudimage: {
+      token: 'demo',
+      apiVersion: 'v7' // optional
```
danielroe, line 63: if this is the default for new tokens we should probably add it above to the example (set to an empty string)

TiborPL: With the breaking changes, I've added a bit more explanation to docs. Should be much clearer how Cloudimage integration works.

docs/content/3.providers/cloudimage.md
```diff
@@ -5,43 +5,83 @@ description: Nuxt Image has first class integration with Cloudimage.
...
+
+### `apiVersion`
+
 - Type: **String**
+- Default: `empty string`
+
+Allow to use a specific version of API.
+
+::callout
+For tokens created before **20.10.2021**, `apiVersion` need to be set to `v7`.
```
danielroe, line 73:
```suggestion
For tokens created before **20.10.2021**, `apiVersion` needs to be set to `v7`.
```

docs/content/3.providers/cloudimage.md
```diff
@@ -5,43 +5,83 @@ description: Nuxt Image has first class integration with Cloudimage.
...
+::
 
-Your uploads base URL. This alows to shorten your origin image URLs.
+Here's an official demo config. `demo` is an old token hence the `apiVersion` need to be defined as well.
```
danielroe, line 76:
```suggestion
Here's an official demo config. `demo` is an old token hence `apiVersion` needs to be defined as well.
```

docs/content/3.providers/cloudimage.md
```diff
@@ -50,7 +90,19 @@ export default defineNuxtConfig({
 ### `cdnURL`
 
 - Type: **String**
-- Default: `https://{token}.cloudimg.io/v7`
+- Default: `https://{token}.cloudimg.io/{apiVersion}`
+
+Replaces the dynamicaly built URL
```
danielroe, line 95:
```suggestion
Replaces the dynamically built URL
```

docs/content/3.providers/cloudimage.md
```diff
@@ -5,43 +5,83 @@ description: Nuxt Image has first class integration with Cloudimage.
...
+      baseURL: 'https://sample.li/images',
+      baseURL: '_sl_' // alias defined in your Cloudimage storage settings
+    }
+  }
+})
+```
+
+
+
+
```
danielroe, line 63:
```suggestion
```

docs/content/3.providers/cloudimage.md
```diff
@@ -5,43 +5,83 @@ description: Nuxt Image has first class integration with Cloudimage.
...
+### `apiVersion`
+
 - Type: **String**
+- Default: `empty string`
+
+Allow to use a specific version of API.
```
danielroe, line 70:
```suggestion
Allow using a specific version of the API.
```

src/runtime/providers/cloudimage.ts
```diff
@@ -21,17 +21,29 @@ const operationsGenerator = createOperationsGenerator({
...
+  apiVersion = '',
   cdnURL = ''
 } = {}) => {
   const operations = operationsGenerator(modifiers)
   if (!cdnURL) {
-    cdnURL = `https://${token}.cloudimg.io/v7`
+    cdnURL = `https://${token}.cloudimg.io/${apiVersion}`
   }
+
+  if (isDev) {
```
danielroe, line 39: This makes it look like this provider will not work in development mode? Should we instead check `isDev && <some other condition>`?

TiborPL: Oh man, I shouldn't do this late night. 
I've added proper condition now.

danielroe (approved):

PR #966, opened 2023-09-05
feat(nuxt-img): placeholder default blur + improve playground

Atinux:
- It renders much better with a default to blur to `3` for the most sizes.
- Improved the playground to add `<NuxtImg>`
- Updated the docs regarding it


pi0 (commented):
Using `blur` seems like a good idea 👍🏼

pi0 (approved):

danielroe (approved):

Issue #912, opened 2023-07-12
Set reasonable cap for density-based srcsets


kara:
High-resolution mobile devices often download images that are larger than necessary due to their high DPR (widths are multiplied by that DPR when selecting an image from the srcset). [Data shows](https://observablehq.com/@eeeps/visual-acuity-and-device-pixel-ratio) that most people stop benefitting from sharpness at about 2 DPR (i.e. can't see a difference at DPRs > 3, etc).  To address this issue, it makes sense to set a cap on srcset density descriptors so apps aren't serving giant images on mobile without realizing it.

Angular [throws an error](https://github.com/angular/angular/commit/b380fdd59e368e89908ea915f150cdc9f5a87a7f) if the image directive detects a custom srcset with a density descriptor higher than 3, e.g. something like:

```
<nuxt-img
  src="/logos/nuxt.png"
  height="50"
  densities="x1 x2 x3 x4"
  />
```

PR #973, opened 2023-09-09
fix: warn in dev mode if density is provided above 2x

danielroe:
resolves https://github.com/nuxt/image/issues/912

This PR warns in dev mode if user is setting a density > 2x, linking to https://observablehq.com/@eeeps/visual-acuity-and-device-pixel-ratio.

It doesn't hard-disable this although this might be something we consider in future.

This also refactors the density parsing expression slightly to use a Set for deduplicating.

### DIFF
--- a/src/runtime/utils/index.ts
+++ b/src/runtime/utils/index.ts
@@ -108,7 +108,7 @@ export function parseDensities (input: string | undefined = ''): number[] {
 
 export function checkDensities (densities: number[]) {
   if (densities.length === 0) {
-    throw new Error('\'densities\' must not be empty, configure to \'1\' to render regular size only (DPR 1.0)')
+    throw new Error('`densities` must not be empty, configure to `1` to render regular size only (DPR 1.0)')
   }
   if (process.dev && Array.from(densities).some(d => d > 2)) {
     console.warn('[nuxt] [image] Density values above `2` are not recommended. See https://observablehq.com/@eeeps/visual-acuity-and-device-pixel-ratio.')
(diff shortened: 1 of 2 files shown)


src/runtime/utils/index.ts
```diff
@@ -97,10 +97,22 @@ export function parseDensities (input: string | undefined = ''): number[] {
...
+}
 
-  // de-duplicate and return
-  return densities.filter((value, index) => densities.indexOf(value) === index)
+export function checkDensities (densities: number[]) {
+  if (densities.length === 0) {
+    throw new Error('\'densities\' must not be empty, configure to \'1\' to render regular size only (DPR 1.0)')
+  }
+  if (process.dev && Array.from(densities).some(d => d > 2)) {
+    console.warn('[nuxt] [image] Density values above 2 are not recommended. See https://observablehq.com/@eeeps/visual-acuity-and-device-pixel-ratio.')
```
pi0, line 114:
```suggestion
    console.warn('[nuxt] [image] Density values above `2` are not recommended. See https://observablehq.com/@eeeps/visual-acuity-and-device-pixel-ratio.')
```

src/runtime/utils/index.ts
```diff
@@ -97,10 +97,22 @@ export function parseDensities (input: string | undefined = ''): number[] {
...
+  }
+
+  return Array.from(densities)
+}
 
-  // de-duplicate and return
-  return densities.filter((value, index) => densities.indexOf(value) === index)
+export function checkDensities (densities: number[]) {
+  if (densities.length === 0) {
+    throw new Error('\'densities\' must not be empty, configure to \'1\' to render regular size only (DPR 1.0)')
```
pi0, line 111:
```suggestion
    throw new Error('`densities` must not be empty, configure to `1` to render regular size only (DPR 1.0)')
```

pi0 (approved):

PR #976, opened 2023-09-11
fix ipx provider options

aaharu:
https://image.nuxt.com/get-started/configuration#modifiers

According to the document, setting to default modifiers should be possible, but it wasn't working in typescript, so I fixed it.


src/ipx.ts
```diff
@@ -28,10 +28,10 @@ export const ipxSetup: ProviderSetup = async (providerOptions, moduleOptions) =>
...
   const ipxOptions = {
     dir: resolve(nuxt.options.srcDir, moduleOptions.dir || nuxt.options.dir.public),
-    maxAge: providerOptions.options?.maxAge,
-    domains: moduleOptions.domains,
-    sharp: moduleOptions.sharp,
-    alias: moduleOptions.alias
+    maxAge: providerOptions.options?.maxAge || moduleOptions.ipx?.maxAge,
+    domains: moduleOptions.domains || moduleOptions.ipx?.domains,
+    sharp: moduleOptions.sharp || moduleOptions.ipx?.sharp,
+    alias: moduleOptions.alias || moduleOptions.ipx?.alias
```
pi0, line 34: New options are updated for ipx v2 interface (https://github.com/nuxt/image/blob/main/src/ipx.ts#L32)

We probably don't need this anymore (but we can also spread entire `moduleOptions.ipx` (typed as new `IPXRuntimeConfig`)

src/types/module.ts
```diff
@@ -78,13 +112,13 @@ export interface ImageProviders {
...
-  imageengine?: any,
-  uploadcare?: Partial<UploadcareOptions>,
-  ipx?: Partial<IPXOptions>
-  static?: Partial<IPXOptions>
+  storyblok?: any
+  wagtail?: any
+  strapi?: any
+  imageengine?: any
+  uploadcare?: Partial<UploadcareOptions>
+  ipx?: Partial<StaticOptions>
```
pi0, line 120: This should be IPXOptions i guess?

src/types/module.ts
```diff
@@ -64,6 +64,40 @@ export interface UploadcareOptions {
...
+  threshold: number | string
+  modulate: string
+  tint: number | string
+  grayscale: true | 'true'
+  animated: true | 'true'
+}
+
+export interface StaticOptions extends Partial<IPXOptions> {
+  baseURL: string
+  modifiers: Partial<IPXModifiers>
```
pi0, line 98: Would be nice to rename it to `StaticIPXOptions` also i think `baseURL` at least is shared also for runtime with v2 migration

src/module.ts
```diff
@@ -139,7 +139,13 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
         imageOptions.provider = options.provider = resolvedProvider
         options[resolvedProvider] = options[resolvedProvider] || {}
 
-        const p = await resolveProvider(nuxt, resolvedProvider, options[resolvedProvider])
+        const p = await resolveProvider(
+          nuxt,
+          resolvedProvider,
+          BuiltInProviders.includes(resolvedProvider)
+            ? { provider: resolvedProvider, options: options[resolvedProvider] }
+            : options[resolvedProvider]
```
pi0, line 147: Was this change necessary? (I think it would be nice to keep this arg unified with `{ options }` interface always.)

aaharu: It's necessary.
The `resolveProviders` that take `ModuleOptions` as arguments also do this processing, so they need to do the same for `ModuleOptions` in the `nitro:init` hook.
https://github.com/nuxt/image/blob/v1.0.0-rc.3/src/provider.ts#L67-L71

pi0: Considering this logic running within `if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic')`, possible values for `resolvedProvider` are `ipx`, `ipxStatic` and `none`, all are built-in so i think this line would not also reach...

pi0:
Hi. I have merged in a slightly different fix via https://github.com/nuxt/image/pull/1056 for options handling (sorry since was in rush for 1.0.0 release).

Would be nice if we can repurpose this changes to modifiers type+fix. (And thanks for all your helps on both nuxt image and IPX all the time ❤️ )

aaharu:
Thank you for reviewing! And sorry for the delay in getting back.

This PR makes some conflicts, so I  remade it following.
https://github.com/nuxt/image/pull/1132

Issue #645, opened 2022-11-11
Allow setting simple view-width based "sizes" values


kara:
To reproduce:
1. Add nuxt/image to modules (in config, yarn add, etc)
1. Add nuxt/image instance with width and height set
1. Add a sizes property based on "vw" units

```
<nuxt-img src="cat.jpg" width="300" height="200" sizes="100vw" />
```

Expected behavior:
Image component has a `srcset` attribute with width-descriptors. The `sizes` attribute falls through to the <img> tag as is. 

Expected result:

```
<img src="cat.jpg" width="300" height="200" 
     srcset="cat.jpg?w=200 200w, cat.jpg?w=400 400w…" 
     sizes="100vw" />
```

Actual behavior:
The image doesn't render. An empty `srcset` attribute and an empty `sizes` attribute is generated. No `src` is generated. 

Actual result:
```
<img width="300" height="200" srcset sizes />
```

Suggested change:
The `sizes` attribute should support passing vw-based values directly to the underlying image tag. A generic `srcset` should be generated with common device sizes.

nomadoda:
Duplicate issue https://github.com/nuxt/image/issues/469

Baroshem:
@danielroe I did some investigation and I have found a solution for the sizes not being generated.

Because there was no sizes, there was also no srcset because it was using it to correctly create a srcset list.

Let me know if that was the idea with this task.

<img width="1159" alt="image" src="https://github.com/nuxt/image/assets/37120330/85244ac8-1ce5-44f2-b1cf-4a9ed71245ed">

PR #977, opened 2023-09-11
feat!: switch `sizes` default to responsive-first

danielroe:
resolves https://github.com/nuxt/image/issues/645

This aims to enable responsive sizing by default. It has two breaking changes:

1. sizes without an explicit screen width (e.g. `sizes="100vw"`) are interpreted as a _responsive_ minimum size, e.g. `1px:100vw`.
2. Sizes are output from that screen-size _up_ rather than that screen-size _down_, so `sizes="md:100px lg:200px"` means that the image is displayed at `100px` by default. And at `lg` and above, at `200px`. (Previously, this meant that it was displayed at 100px only up to `md`, and then displayed at `200px` for all sizes above that.)

As these are both significant changes from previous behaviour, special care is requested in evaluating the API and DX.

### DIFF
--- a/docs/content/2.usage/1.nuxt-img.md
+++ b/docs/content/2.usage/1.nuxt-img.md
@@ -55,15 +55,20 @@ Specify width/height of the image.
 
 Specify responsive sizes.
 
-This a space-separated list of screen size/width pairs. You can [see a list of the defined screen sizes here](/get-started/configuration#screens)).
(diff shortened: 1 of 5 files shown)


Atinux:
It looks good to me.

What changes are required in the docs?

danielroe:
Will update :+1:

pi0 (approved):
overall looks a good idea to base on smaller sizes 👍🏼

PR #985, opened 2023-09-14
fix(cloudinary): handle blur modifier correctly

### LINKED ISSUE #983: Placeholder creates bad request on cloudinary
Hey everyone,

Thanks for your amazing work ❤️

I am using: 
Version: @nuxt/image": "1.0.0-**rc.2**"
Provider: **cloudinary**

I get a 400 (bad request) as soon as I add the `placeholder` prop to the `<nuxt-img />` component.
e.g.
```vue
<nuxt-img
  ...
  placeholder
/>
```

This is the generated request:
> https://res.cloudinary.com/df5skrrwb/image/upload/f_webp,q_50,w_10,h_10,blur_3/v1677729457/travel/taiwan/taipei_sunsetview_edited_diwwcg.png

The [cloudinary docs](https://cloudinary.com/documentation/transformation_reference#e_blur) specify a different syntax for _blur_: 
`e_blur[:<strength>]` instead of `blur_[:<strength>]`

If I apply _blur_ as documented, the generated link works as expected:
> https://res.cloudinary.com/df5skrrwb/image/upload/f_webp,q_50,w_10,h_10,e_blur:3/v1677729457/travel/taiwan/taipei_sunsetview_edited_diwwcg.png

## Further information
This bug was introduced with https://github.com/nuxt/image/pull/966

PatrikBird:
This fixes #983

### DIFF
--- a/src/runtime/providers/cloudinary.ts
+++ b/src/runtime/providers/cloudinary.ts
@@ -74,7 +74,7 @@ const operationsGenerator = createOperationsGenerator({
     }
   },
   joinWith: ',',
-  formatter: (key, value) => key === 'e_blur' ? `${key}:${value}` : `${key}_${value}`
+  formatter: (key, value) => key.includes('_') ? `${key}:${value}` : `${key}_${value}`
 })
 
 const defaultModifiers = {
--- a/src/runtime/providers/cloudinary.ts
+++ b/src/runtime/providers/cloudinary.ts
@@ -29,7 +29,8 @@ const operationsGenerator = createOperationsGenerator({
     colorSpace: 'cs',
     customFunc: 'fn',
     density: 'dn',
-    aspectRatio: 'ar'
+    aspectRatio: 'ar',
+    blur: 'e_blur'
   },
   valueMap: {
     fit: {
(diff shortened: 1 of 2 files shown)


src/runtime/providers/cloudinary.ts
```diff
@@ -73,7 +74,7 @@ const operationsGenerator = createOperationsGenerator({
     }
   },
   joinWith: ',',
-  formatter: (key, value) => `${key}_${value}`
+  formatter: (key, value) => key === 'e_blur' ? `${key}:${value}` : `${key}_${value}`
```
PatrikBird, line 77: I _feel_ dirty suggesting this. 😬 
I am not sure how this is supposed to be like as cloudinary has [a lot of transformations](https://cloudinary.com/documentation/transformation_reference#e_blur) with different formats e.g. `e_transformation:<value>` / `w_<value>` or without any value at all.
- should all be supported or is `e_blur` enough in this case?

Baroshem: I would say that this is the way for doing it. Take a look at https://github.com/colbyfayock/cloudinary-util/blob/700863c02a18712e0ce01b8e7511449b21e40dc6/packages/url-loader/src/lib/transformations.ts#L35. This is currently the most supported cloudinary url loader and it uses the same approach. 

As there are other params like blur, I would recommend to changing it to something like this:

```suggestion
  formatter: (key, value) => key.includes('_') ? `${key}:${value}` : `${key}_${value}`
```

Baroshem (commented):
Hey, very good Pr, please take a look at the comment I added :)

Baroshem (approved):

danielroe (approved):

PR #988, opened 2023-09-14
feat!: migrate to ipx v2 

pi0:
Migrate to IPX v2 for better future compatibility 🚀 

v2 roadmap: https://github.com/unjs/ipx/issues/171

- [x] Update types
- [x] Working dev handler
- [x] Working runtime -- logic is now unified with runtime
- [ ] ~~Integrate with storage layer for cache support~~ (deferred)


src/runtime/ipx.ts
```diff
@@ -1,23 +1,29 @@
...
+  const fsDir = opts.fs?.dir ? isAbsolute(opts.fs.dir) ? opts.fs.dir : fileURLToPath(new URL(opts.fs.dir, import.meta.url)) : undefined
+
+  const fsStorage = opts.fs?.dir ? ipxFSStorage({ ...opts.fs, dir: fsDir }) : undefined
+  const httpStorage = opts.http?.domains ? ipxHttpStorage({ ...opts.http }) : undefined
+  if (!fsStorage && !httpStorage) {
+    throw new Error('IPX storage is not configured!')
+  }
+
+  const ipxOptions: IPXOptions = {
+    storage: (fsStorage || httpStorage)!,
```
Atinux, line 21: This means we can only use on storage? Either FS or http?

src/types/module.ts
```diff
@@ -1,9 +1,10 @@
 import type { IPXOptions } from 'ipx'
+import type { Nuxt } from 'nuxt/schema'
```
danielroe, line 2:
```suggestion
import type { Nuxt } from '@nuxt/schema'
```

test/e2e/generate.test.ts
```diff
@@ -33,11 +38,9 @@ describe('ipx provider', () => {
         "/_ipx/s_300x300/images/colors.jpg",
         "/_ipx/s_300x300/images/everest.jpg",
         "/_ipx/s_300x300/images/tacos.svg",
-        "/_ipx/s_300x300/unsplash/photo-1606112219348-204d7d8b94ee",
```
danielroe, line 36: @pi0 shouldn't we be generating files for remote sources as well when prerendering?

src/runtime/ipx.ts
pi0: It will be soon unstorage by default in next iteraitons so pluggable and also we can replace it with chainable storages

danielroe:
having talked to @pi0, marking as draft until we address https://github.com/nuxt/image/pull/988#discussion_r1334139294.

test/e2e/generate.test.ts
pi0: found the issue alias was not being passed

danielroe (approved):

Issue #944, opened 2023-08-24
IPX provider is breaking image source query parameters resulting in failing SSG


Predota:
### Issue :
Instagram is protecting their media URLs with query parameters. 
When using nuxt generate, IPX is encoding the source breaking the query parameters in the process.

### Nuxt project info :
------------------------------
- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.6.0
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      modules, image
- Runtime Modules:  @nuxt/image@1.0.0-rc.1
- Build Modules:    -
------------------------------

### Reproduction :
https://stackblitz.com/edit/github-mef1ym?file=nuxt.config.ts,app.vue,package.json

Issue #815, opened 2023-04-27
IPX returns a '403: Fetch error' during static generation for image urls contai special characters


arthurmchr:
nuxt: 3.4.0
@nuxt/image-edge: 1.0.0-28020728.5df24eb

In my `nuxt.config.ts` I have :
```
modules: ['@nuxt/image-edge'],
image: {
  domains: ['cdn.discordapp.com'],
}
```

In my `app.vue` file I only have the following :
```
<template>
  <div>
    <nuxt-img src="https://cdn.discordapp.com/my-image.png?test=ok"
    />
  </div>
</template>
```

It works well in development.

But when I execute the `npm run generate` command to export a static version of my site, during the prerendering, I get
```
[nitro 11:52:32]   ├─ /_ipx/_/https://cdn.discordapp.com/my-image.png%3Ftest=ok (205ms) (Error: [403] IPX: Fetch error)
```
The error appears because of the replacement of the `?` character in my URL by `%3F`.
I will get the same error for all file with a ` `(SPACE) character in the name, or special characters in it.
I got the same result on other domain too (not only `cdn.discordapp.com`)

arthurmchr:
Removing `encodePath` in [https://github.com/nuxt/image/blob/main/src/runtime/providers/ipx.ts](https://github.com/nuxt/image/blob/main/src/runtime/providers/ipx.ts) (line 33) seems to fix it.

maartenvanasperen:
> Removing `encodePath` in https://github.com/nuxt/image/blob/main/src/runtime/providers/ipx.ts (line 33) seems to fix it.

I got the same issue, and removing the `encodePath` seems to fix it. It only occurs with ssg and not with a dev build

McSundae:
I have the exact same problem, but i dont think it would be right to make changes in those files to fix this problem. There needs to be some sort of option to fix that problem that we are just missing out...

PR #997, opened 2023-09-19
fix(ipx): fix retrieving images with query params 

matthijsch:
**Problem**
images containing query params cannot be retrieved in IPX because the query params are encoded

_issues_
https://github.com/nuxt/image/issues/815
https://github.com/nuxt/image/issues/944

**Solution**
by using ufo's normalizeURL the src is split up in host, pathname, query etc before any encoding is done. This way only the path name is encoded

_before_
```/_ipx/w_500%26%26q_80/https://host.com/uploads/image%20file.jpg%3Fv=1%26foo=bar```                                                                                                                                   
_after_
```/_ipx/w_500%26%26q_80/https://host.com/uploads/image%20file.jpg?v=1&foo=bar```

**Remaining problem**
This solves retrieving images, what remains to be fixed is writing the image to the disk as Nitro prevent saving images with routes containing "?" since 2.6.0:
https://github.com/unjs/nitro/pull/1474


pi0:
Thanks for this PR. 

(forwarding from discord) 

While this solution might work to solve encoding issues it seems not safer. Adding URLs to the ends of a URL is already not totally safe for parsers and UFO normalize util tends to avoid extra optional encodings (following the vue-router approach). By doing this change, we can potentially introduce new edge cases.

We might investigate the upstream ipx server to handle double encoding.

I will try to follow up on this soon.

danielroe:
@pi0 Did you end up resolving upstream in `ipx`? I'm aware there's been a major release since this comment so I think likely.

barayuda:
Hi everyone @pi0 @danielroe any update regarding this?
I got this error as well when do SSG

TechAkayy:
Hi @pi0 @danielroe 🙂,

Can this PR be merged? It inherently fixes a key bug with nuxt generate when optimizing remote images with params. The params are getting double-encoded and break static site generation. Currently, the only workaround seems to be falling back to using the  tag.

I recently noticed some heated debate on Reddit - https://www.reddit.com/r/Nuxt/comments/1e7t9dp/is_nuxt_fit_for_static_sites_or_astro/?utm_source=share&utm_medium=mweb3x&utm_name=mweb3xcss&utm_term=1&utm_content=share_button.

The optimization to prevent this double-encoding in ipx could be pursued separately.

This probably fixes a few existing issues with nuxt generate. Unless I'm mistaken, please correct me if you see any regression with this normalization.

Thanks 🙏🏾 bunch!

Issue #1011, opened 2023-09-27
Support for nonce or hash value [enhancment]


rahul37865:
Recently I added Nuxt Security in a Project where i am fetching data from an API URL http://127.0.0.1:8000/api/blog/post/27
i am implementing Nuxt Image Module and While displaying image 
`<NuxtImg crossorigin="anonymous" :src="postDetail.post_image" />`
It throws an error (only on Firefox, doesn't throw error on chrome and edge)
```
Content-Security-Policy: The page’s settings blocked the loading of a resource at inline (“script-src-attr”).
Source: this.setAttribute('data-error', 1)
```
which i fixed using below settings
  ```
security: {
    headers: {
      contentSecurityPolicy: {
        'img-src': ["'self'", 'data:', 'http://127.0.0.1:8000'],
        'script-src-attr': ["'unsafe-inline'"],        // Added this
      },
    },
  },
```
But it is considered a security risk, as it can open the door to XSS attacks. It's generally recommended to avoid using 'unsafe-inline' whenever possible. I discussed this matter with Nuxt Security and they recommend to use nonce or hash value. They also said that unfortunately Nuxt Image doesnt support nonce or hash so there is no secure way. You can find the discussion here https://github.com/Baroshem/nuxt-security/issues/218

trijpstra-fourlights:
I've created a reference implementation on the `nonce` propagation here https://github.com/trijpstra-fourlights/nuxt-image/tree/feat/add-nonce-support

I'll see if I can add the `hash` support as well and create a proper PR for it.

Baroshem:
I am really looking forward to this feature as it would enable working well with Nuxt Image & Nuxt Security :)

Baroshem:
Hey @pi0 could you take a look at it? It would be really useful to have this feature. Thomas is doing an amazing work in the development of Nuxt Security features and now prepared a feature for Nuxt Image that would help not only in the case of Nuxt security, but also for other projects as well :)

rahul37865:
Yes Team Please do check and if everything looks good then proceed to merge.

Baroshem:
@pi0 

I think this issue could be closed as the PR was merged :)

dargmuesli:
I don't think that this issue is resolved fully yet. The `this.setAttribute('data-error', 1)` inline script is still inline `onerror` and results in the given error. It appears to me that the nonce is currently only added to the preload header.

P4sca1:
[...]

PR #1012, opened 2023-09-27
feat: add `nonce` support

trijpstra-fourlights:
Implements #1011

### DIFF
--- a/docs/content/2.usage/1.nuxt-img.md
+++ b/docs/content/2.usage/1.nuxt-img.md
@@ -278,7 +278,9 @@ Providing a nonce allows you to avoid using the CSP `unsafe-inline` directive, w
 <NuxtImg src="/nuxt-icon.png" :nonce="nonce" />
 
 <script lang="ts" setup>
-const nonce = useNonce() // using `nuxt-security` module
+// useNonce is not provided by nuxt/image but might be
+// provided by another module, for example nuxt-security
+const nonce = useNonce()
 </script>
 ```
 
--- a/docs/content/2.usage/1.nuxt-img.md
+++ b/docs/content/2.usage/1.nuxt-img.md
@@ -269,6 +269,20 @@ Set `loading="lazy"` to defer loading of an image until it appears in the viewpo
 <NuxtImg src="/nuxt-icon.png" loading="lazy" />
 ```
 
+### `nonce`
+
+This is a [native](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) global attribute that defines a cryptographic nonce (number used once) that can be used by Content Security Policy to determine whether or not a given fetch will be allowed to proceed for a given element.
+Providing a nonce allows you to avoid using the CSP `unsafe-inline` directive, which would allowlist _all_ inline script or styles.
+
+```html
+<NuxtImg src="/nuxt-icon.png" :nonce="nonce" />
+
+<script lang="ts" setup>
+const nonce = useNonce() // using `nuxt-security` module
+</script>
+```
+
+```html
 
 ## Events
 
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -79,7 +79,8 @@ export default defineComponent({
         link: [{
           rel: 'preload',
           as: 'image',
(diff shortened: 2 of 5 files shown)


docs/content/2.usage/1.nuxt-img.md
```diff
@@ -269,6 +269,20 @@ Set `loading="lazy"` to defer loading of an image until it appears in the viewpo
...
+### `nonce`
+
+This is a [native](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) global attribute that defines a cryptographic nonce (number used once) that can be used by Content Security Policy to determine whether or not a given fetch will be allowed to proceed for a given element.
+Providing a nonce allows you to avoid using the CSP `unsafe-inline` directive, which would allowlist _all_ inline script or styles.
+
+```html
+<NuxtImg src="/nuxt-icon.png" :nonce="nonce" />
+
+<script lang="ts" setup>
+const nonce = useNonce() // using `nuxt-security` module
```
danielroe, line 281:
```suggestion
// useNonce is not provided by nuxt/image but might be
// provided by another module, for example nuxt-security
const nonce = useNonce()
```

src/runtime/components/nuxt-img.ts
```diff
@@ -78,6 +78,7 @@ export default defineComponent({
       useHead({
         link: [{
           rel: 'preload',
+          nonce: props.nonce,
```
danielroe, line 81:
```suggestion
          ...(props.nonce ? { nonce: props.nonce } : {}),
```

trijpstra-fourlights: I see you removed this quite quickly after my recent push in https://github.com/nuxt/image/pull/1012/commits/d1d5258b9711f8669b70ef94e5c1d32987e4d90b 

Locally `pnpm test:types` (on https://github.com/nuxt/image/pull/1012/commits/07b141e0d82b742ed81f908911af898a819bee48) runs without error (and without the need for the `ts-expect-error`).

danielroe: sorry, I was editing locally at the same time 🙈

the reason I prefer this is that the spread _does_ work without type errors, but it also removes type safety here as you could pass anything in, e.g. `{ bob: true }`.

trijpstra-fourlights: I see, makes sense! Thanks for all your hard work on nuxt btw :)

Baroshem (approved):

src/runtime/components/nuxt-picture.ts
```diff
@@ -1,4 +1,5 @@
 import { h, defineComponent, ref, computed, onMounted } from 'vue'
+import type { Head } from 'zhead'
```
harlan-zw, line 2: zhead is an implicit dependency, better to use the explicit dependency 

```ts
import type { Head } from '@unhead/vue'
```

src/runtime/components/nuxt-img.ts
```diff
@@ -79,6 +79,8 @@ export default defineComponent({
         link: [{
           rel: 'preload',
           as: 'image',
+          // @ts-expect-error might need to resolve upstream in unhead
```
harlan-zw, line 82: This is fixed in zhead [v2.1.2](https://github.com/harlan-zw/zhead/releases/tag/v2.1.2)

trijpstra-fourlights: In the context of this PR, what's the preferred resolution here? 

Wait for the new release of `@unhead/vue` to use `zhead@^2.1.2`? Or do you want me to update the comment here to something like `fixed in zhead v2.1.2, needs upstream adoption by @unhead/vue`?

harlan-zw: Good question, if you delete the lock file & node_modules it should update to this version and you can remove the comment altogether

danielroe: don't worry about refreshing lockfile in this pr; I can remove the comment when we next update deps....

danielroe (approved):

harlan-zw (approved):

Baroshem:
Awesome! 🚀

Issue #1026, opened 2023-10-03
feat: baseURL must be modified when src start by http(s)


joffreyBerrier:
I would like to have a baseUrl by default, but when src starts by HTTP or HTTPS, I would like not to use the baseUrl and use the src instead, which keeps params like width, height, etc because I'm using two CDN but the cdn always use `cloudimage`

In my **nuxt.config.ts**
```ts
image: {
  provider: 'cloudimage',
  cloudimage: {
    token: '****',
    baseURL: '__path__',
    format: ['webp'],
  },
  quality: 65,
},
```

In my **components**
```vue
<script setup lang"ts">
const src = "https://cdn.cloudimage.com/image/1.jpg"
</script>

<template>
  <NuxtImg
    fit="cover"
    alt="alt"
    densities="x1 x2"
    :src="src"
    height="100"
    width="100"
    loading="lazy"
  />
</template>
```

**Today **NuxtImg** render this :**
`https://token.cloudimg.io/__path__/https://cdn.cloudimage.com/image/1.jpg?height=100&width=100&q=65&func=crop`

**And not this:**
`https://placehold.co/600x400?height=100&width=100&q=65&func=crop`

**Pull request**
https://github.com/nuxt/image/pull/1028

PR #1028, opened 2023-10-03
fix(cloudimage): skip cdn when `src` path is provided with protocol

joffreyBerrier:
**Problem**
* **Cloudimage**: Images starting with `HTTP` should not be prefixed with `Base Url`

**issues**
* https://github.com/nuxt/image/issues/1026

**Solution**
* Add condition if src `startwith http` return src with generate query parameters

**Before:**
* `https://token.cloudimg.io/__path__/https://2412819702-files.gitbook.io/~/files/v0/b/gitbook-x-prod.appspot.com/o/spaces%2FlIgyYELwJG6odLEyCM6i%2Fuploads%2FAHcbuKRYbIlBWO4cJ88b%2Fimage.png?alt=media&token=62ff753d-83eb-4e3f-932c-96eb72d455f1?height=100&width=100&q=65&func=crop`

**After:**
* `https://2412819702-files.gitbook.io/~/files/v0/b/gitbook-x-prod.appspot.com/o/spaces%2FlIgyYELwJG6odLEyCM6i%2Fuploads%2FAHcbuKRYbIlBWO4cJ88b%2Fimage.png?alt=media&token=62ff753d-83eb-4e3f-932c-96eb72d455f1?height=100&width=100&q=65&func=crop`

**Remaining**
* I have tried to do that for all providers, but the provider manages the logic of the URL, so I can't test this and can't do that for all providers. Do you have any idea?

### DIFF
--- a/src/runtime/providers/cloudimage.ts
+++ b/src/runtime/providers/cloudimage.ts
@@ -55,6 +55,12 @@ export const getImage: ProviderGetImage = (src, {
     cdnURL = `https://${token}.cloudimg.io/${apiVersion}`
   }
 
+  if (src.startsWith('http')) {
+    return {
+      url: joinURL(src) + (operations ? ('?' + operations) : '')
+    }
+  }
+
   return {
     url: joinURL(cdnURL, baseURL, src) + (operations ? ('?' + operations) : '')
   }
--- a/src/runtime/providers/cloudimage.ts
+++ b/src/runtime/providers/cloudimage.ts
@@ -1,4 +1,4 @@
-import { joinURL } from 'ufo'
+import { joinURL, hasProtocol } from 'ufo'
 import type { ProviderGetImage } from '../../types'
 import { createOperationsGenerator } from '#image'
 
@@ -30,6 +30,7 @@ export const getImage: ProviderGetImage = (src, {
   cdnURL = ''
 } = {}) => {
   const operations = operationsGenerator(modifiers)
+  const query = (operations ? ('?' + operations) : '')
 
   if (process.dev) {
     const warning = []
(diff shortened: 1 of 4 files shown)


src/runtime/providers/cloudimage.ts
```diff
@@ -55,6 +55,12 @@ export const getImage: ProviderGetImage = (src, {
...
 
+  if (src.startsWith('http')) {
+    return {
+      url: joinURL(src) + (operations ? ('?' + operations) : '')
+    }
+  }
+
   return {
     url: joinURL(cdnURL, baseURL, src) + (operations ? ('?' + operations) : '')
   }
```
danielroe, line 66: I think we can slightly simplify this:

```suggestion
  const query = (operations ? ('?' + operations) : '')

  if (src.startsWith('http')) {
    return {
      url: src + query
    }
  }

  return {
    url: joinURL(cdnURL, baseURL, src) + query
  }
```

src/runtime/providers/cloudimage.ts
```diff
@@ -55,6 +55,12 @@ export const getImage: ProviderGetImage = (src, {
     cdnURL = `https://${token}.cloudimg.io/${apiVersion}`
   }
 
+  if (src.startsWith('http')) {
```
danielroe, line 58: Can we perhaps use a more precise test? for example, this would match `http-some-file.png`.

danielroe (changes requested):
Thank you ❤️ 

I don't think we need to update other providers in this PR.

Could you update the check on `src` to make it a bit more precise? 🙏

danielroe (approved):
this looks good to me, thank you 👌

danielroe (approved):
this looks good to me, thank you 👌

Atinux (approved):

PR #1029, opened 2023-10-03
docs: add title to provider links

Lehoczky:
## 📚Description

Just a small update to add the native browser tooltip for the providers in the home page. This way people who don't recognize a particular logo can just hover over the icons and see the provider's name. 

image


Atinux:
Nice catch thank you @Lehoczky

PR #1066, opened 2023-10-20
feat: add `sirv` provider

IgorVaryvoda:
This PR adds support for [Sirv](https://sirv.com/help/articles/dynamic-imaging/) image provider


src/runtime/providers/sirv.ts
```diff
@@ -0,0 +1,137 @@
...
+    format: 'format',
+    webpFallback: 'webp-fallback',
+    subsampling: 'subsampling',
+    gifCompression: 'gif.lossy',
+    crop: 'crop.type',
+    cropAr: 'crop.aspectratio',
+    cw: 'cw',
+    ch: 'ch',
+    cx: 'cx',
+    cy: 'cy',
```
danielroe, line 38: you can omit the keys that are equal to their values; the generator will just return the key if there's not an entry in the key map.

https://stackblitz.com/edit/github-z7xwz3?file=nuxt.config.ts,app.vue

danielroe (commented):
This looks great :heart:

src/runtime/providers/sirv.ts
IgorVaryvoda: I see, thanks, addressed in https://github.com/nuxt/image/pull/1066/commits/d876277f5585b78b220ed68d0ce7c2b0c1c41cdf

danielroe (approved):

PR #1124, opened 2023-11-23
docs: explain fall back to larger size on amplify + vercel

korve:
The documentation states that the image provider will fallback to the >nearest< image width when a image width was not defined in the screens option. This is not the actual behavior. Actually, the next bigger screen width is return. I've tried to reflect this with this change.


pi0:
Thanks! Would you mind to also reflect same change for `aws-amplify`? 🙏🏼

korve:
> Thanks! Would you mind to also reflect same change for `aws-amplify`? 🙏🏼

Yep, done :)

danielroe (approved):

PR #1132, opened 2023-11-29
chore: add types to ipx provider options

aaharu:
rework of https://github.com/nuxt/image/pull/976

> https://image.nuxt.com/get-started/configuration#modifiers
> According to the document, setting to default modifiers should be possible, but it wasn't working in typescript, so I fixed it.


Atinux (approved):

src/types/module.ts
```diff
@@ -65,6 +65,39 @@ export interface UploadcareOptions {
...
+  negate: true | 'true'
+  normalize: true | 'true'
+  threshold: number | string
+  modulate: string
+  tint: number | string
+  grayscale: true | 'true'
+  animated: true | 'true'
+}
+
+export interface StaticIPXOptions extends Omit<IPXRuntimeConfig, 'alias'> {
```
pi0, line 97: Since after refactors i guess simply call interface IPXOtions..

src/types/module.ts
```diff
@@ -65,6 +65,39 @@ export interface UploadcareOptions {
   [key: string]: any
 }
 
+export interface IPXModifiers extends ImageModifiers {
```
pi0, line 68: Would you please add a comment above to ipx docs and if had time open an issue upstream to remember adding operation types from upstream? 🙏

src/types/module.ts
```diff
@@ -85,8 +118,8 @@ export interface ImageProviders {
   strapi?: any,
   imageengine?: any,
   uploadcare?: Partial<UploadcareOptions>,
-  ipx?: Partial<IPXRuntimeConfig>
-  static?: Partial<IPXRuntimeConfig>
+  ipx?: Partial<StaticIPXOptions>
+  static?: Partial<StaticIPXOptions>
```
danielroe, line 122:
```suggestion
  ipx?: Partial<IPXOptions>
  static?: Partial<IPXOptions>
```

src/types/module.ts
danielroe, line 97:
```suggestion
export interface IPXOptions extends Omit<IPXRuntimeConfig, 'alias'> {
```

src/types/module.ts
pi0, line 68:
```suggestion
// Reference: https://github.com/unjs/ipx?tab=readme-ov-file#modifiers
// TODO: https://github.com/unjs/ipx/issues/199
export interface IPXModifiers extends ImageModifiers {
```

pi0 (approved):

PR #1138, opened 2023-12-01
ci: fix release job conditional

BobbieGoede:
This fixes the release job conditional so it only checks the merge commit title and not the squashed commits in the commit body. This could incorrectly skip/prevent a release if the body contains one of the checked keywords.

I'm opening this PR in multiple repositories, in this case since it only checks for `'[skip-release]'` so it's much less likely to be triggered by squashed commit messages. Still opening this PR for consistency 😅


BobbieGoede:
Closed as this only checks for `'[skip-release]'` which is never (or shouldn't be) at the start of a commit message

PR #1141, opened 2023-12-06
docs: added deprecated message Netlify Large Media

MLoth:
**Just a minor thing!**
Netlify Large Media is deprecated. I added a small inline warning at the top of the documentation with some remarks from the official documentation. I linked to the official forum for a more in depth explanation.

I was wondering if this should also be mentioned in the title?
You could also argue that 'Netlify' should be hidden from the 'providers' navigation?

### DIFF
--- a/docs/content/3.providers/netlify.md
+++ b/docs/content/3.providers/netlify.md
@@ -8,7 +8,9 @@ links:
     size: xs
 ---
 
-> ❗ Netlify’s Large Media service is [deprecated](https://answers.netlify.com/t/large-media-feature-deprecated-but-not-removed/100804). If this feature is already enabled, Large Media will continue to work on these sites as usual. New Large Media configuration is not recommended.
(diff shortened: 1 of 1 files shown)


docs/content/3.providers/netlify.md
```diff
@@ -8,6 +8,8 @@ links:
     size: xs
 ---
 
+> ❗ Netlify’s Large Media service is [deprecated](https://answers.netlify.com/t/large-media-feature-deprecated-but-not-removed/100804). If this feature is already enabled, Large Media will continue to work on these sites as usual. New Large Media configuration is not recommended.
```
Atinux, line 11:
```suggestion
::callout{color="orange" icon="i-ph-warning-duotone"}
Netlify’s Large Media service is [deprecated](https://answers.netlify.com/t/large-media-feature-deprecated-but-not-removed/100804). If this feature is already enabled, Large Media will continue to work on these sites as usual. New Large Media configuration is not recommended.
::
```

PR #1154, opened 2023-12-17
feat: add `caisy` provider

timbenniks:
I made this one a while back. Why not contribute it to the repo? Let's do it!


timbenniks:
@danielroe it looks like some older commits are also shown in this PR. I only want that last one to be added. Anything I should do differently?

danielroe (approved):
Nice! Thank you ❤️ 

(Next time you can use a branch, which will prevent the commits from your previous PRs sticking around.)

Issue #940, opened 2023-08-16
Module doesn't work with nuxt layer images


dimaTarhan:
Hi there. Thanks for cool module.
I use "nuxt-layers" with images in public folder. Also, I use default config for "nuxt-image" module for my test. 
Everything is good with main project images, they loaded correct. 

![image](https://github.com/nuxt/image/assets/43604553/7e2c089c-342d-42fa-9b25-1a1b6ccd563d)

As for images from layer, there are problems. I think, this images don't include in build process in time...

![image](https://github.com/nuxt/image/assets/43604553/dc9740b7-6eb2-4766-8c7f-caa3d2138d68)

Could you give me some recommendations about that?

oemer-aran:
I have the same issue. Images in the`/public` folder of a nuxt layer will cause 404 error for any images you want to load with `nuxt-img`.

Let's say Layer 2 extends Layer 1. Layer 1 has a `/public/image.png`. 

Using `nuxt-img `:
```html
<!-- response will be 404 -->
<nuxt-img src="/image.png"/>
```

Using plain html:
```html
<!-- response is 200 -->
<img src="/image.png"/>
```

Basically `@nuxt/image` should account for images from other layers.

oemer-aran:
I fixed it by adding this to the base layer / `layer-1`:
```ts
// layer-1/nuxt.config.ts
import { createResolver } from "@nuxt/kit"
const { resolve } = createResolver(import.meta.url)

export default defineNuxtConfig({
  image: {
    dir: resolve("./public"),
  },
})
```
```ts
// ./layer-2/nuxt.config.ts
export default defineNuxtConfig({
  extends: ["../layer-1"],
})
```

But this approach will make the `/public` folder in the current layer not work then (`layer-2/public`).

darthf1:
@danielroe apologies for pinging you; is there something we can do to fix this? So moving `/public` to layers does not work for the `nuxt-image` component, but works for the regular `<img>` tag. If you can give some pointers I can try a PR.

danielroe:
@darthf1 Here would be my debugging steps:

1. identify it is in dev or build mode or both
2. check what IPX is doing - is it performing a network request to get the image, or getting it from the FS directly?
3. see what path it's trying to resolve if it's from the FS, or try the network request yourself, manually
4. fix the bug

Aareksio:
Attaching reproduction:
https://stackblitz.com/edit/nuxt-image-ph6pdk

[...]

PR #1177, opened 2024-01-04
feat(ipx): support nuxt layers

Aareksio:
Issue: #940

The issue is caused by IPX receiving only the root `public` directory. IPX's `node-fs` storage does not support multiple directories. I've migrated to `unstorage` (as per TODO left in comment. Next I implemented `combine` driver which combines multiple unstorage storages, at least the parts which gets used by IPX.

---
### Considerations

1. The performance of `combine` driver is probably a bit worse than if it was all single storage, but I do not think this is a big issue considering it gets enabled only in dev, when using layers with public assets. And even then, I do not think any degradation would be noticeable with sane number of images.

2. Please notice `unstorage` added to `devDependencies`. I am unsure where it belongs. `ipx` is `devDependency` and is marked as `external` in `build.config.ts`. But I never had to install `ipx` separately when adding `@nuxt/image`, not quite sure how it works.

3. `IPXRuntimeConfig` has changed. Please check if it makes sense. The is only one `fs-lite` driver option that could be exposed (`ignore`). As the directory is supposed to be deployed (public), I do not think there is a point in allowing `@nuxt/image` user configuring underlying storage solution. Ignoring should be either dealt on module level or IPX. Not in provider configuration.

4. I find the `image.dir` module option weird in context of `ipx` code. In `module.ts` we resolve it to absolute path, but later in `ipx.ts` we resolve it again...? This also makes it impossible to resolve against layers. Docs have 3 notes explaining contexts the option does **not** work. Maybe it is a good idea to remove it and just recommend people to stick to `public`?

---

@pi0 I could not figure out how to use `unstorageToIPXStorage` from `unstorage`. It requires `prefix` and does not translate `/` to `:` in path, making it unusable. It is probably a good idea to update `unstorage` with proper implementation and remove the makeshift one from here. Also a proper built-in `combine` utility, or `node-fs-multiple` driver would have helped a lot.


src/runtime/ipx.ts
```diff
@@ -1,19 +1,116 @@
...
+import fsLiteDriver from 'unstorage/drivers/fs-lite'
+import type { IPXRuntimeConfig } from '../ipx'
 import { useRuntimeConfig } from '#imports'
 
+// Storage and Driver are not exported from unstorage
+type StorageOrDriver = Parameters<typeof originalUnstorageToIPXStorage>[0]
+
+// Copied from ipx source, removed prefix and fixed to work with unstorage keys
+// Ref: https://github.com/unjs/ipx/blob/main/src/storage/unstorage.ts
+export function unstorageToIPXStorage (
```
pi0, line 17: This is nice but I'm wondering we could directly support multiple dirs for IPX to swerve  to avoid all this logic that could also be potentially out of sync.

pi0: (ipx does have it's own fs driver btw)

Aareksio: - `unstorageToIPXStorage` needs to be fixed in IPX
- `ipxFSStorage` can be made to support multiple directories

Both are _better_ solutions. `unstorageToIPXStorage` needs to be fixed, as the current implementation does not work. However, main focus of this PR is squashing a bug from attached issue. Layers are niche use case, I think it should be handled locally.

When deciding to externalise solution, I propose considering `multi-fs` driver for `unstorage` instead of baking another driver into `ipx`.

pi0: We shall simply support array for current IPX fs-provider. unstorage support for IPX is experimental only.

Aareksio: How do you see the support for multiple dirs? Similar to how it's done with `combineUnstorages` here? Attempt to read file from every dir in series?

pi0: Exactly. We shall iterate over (sanetized) dirs simply for when trying to load/stat a file instead of checking one probably also with a refactor to simplify stat check in a util.

Aareksio:
@pi0 Migratated the PR to use unjs/ipx#203. Tested locally with ipx@2.1.0, I did not want to update deps as part of feature, assuming they gonna be bumped independently. I've also tried to mimic the style from your commits on my other PRs, minimal diff, helper functions after main one, do not break logic if not necessary. Hope it fits your vision :)

Also kept the TODO comment, since we reverted back to ipx's fs driver.

danielroe (approved):
Amazing ❤️ 🔥

Aareksio:
Sorry for not pushing the update myself, got caught in other work 😅

danielroe:
Not to worry! This is a great improvement. ❤️

src/ipx.ts
```diff
@@ -66,6 +67,26 @@ export const ipxSetup: IPXSetupT = setupOptions => (providerOptions, moduleOptio
   }
 }
 
+async function getDevDirs (nuxt: Nuxt, moduleOptions: Parameters<ProviderSetup>[1]) {
+  let fs: { existsSync: (path: string) => boolean }
+  try {
+    // IPX is available in node context only
+    // Therefore it should be safe to require fs
+    fs = await import('node:fs')
```
pi0, line 75: This code lives in builder namespace. You can safely import fs in top level of module

src/ipx.ts
```diff
@@ -27,7 +28,7 @@ export const ipxSetup: IPXSetupT = setupOptions => (providerOptions, moduleOptio
   }
 
   // Options
-  const absoluteDir = resolve(nuxt.options.srcDir, moduleOptions.dir || nuxt.options.dir.public)
+  const absoluteDir = await getDevDirs(nuxt, moduleOptions)
```
pi0, line 31: nit pick: `absoluteDirs` or `publicDirs` or `publicDevDirs` could be better new names for this variable.

pi0 (approved):
Other than two above comments looks nice thanks for helping on this ❤️

PR #1181, opened 2024-01-05
Vercel: Removed restriction to use only predefined width's

### LINKED ISSUE #1180: Vercel: Predefined sizes should not use `image.screens`
I am trying to simply resize an image:

```
<NuxtImg src="/img/my-image.png" sizes="400 md:600 lg:800" />
```

This works fine on dev where `ipx` is used. Deployed to Vercel however it uses the provider `'vercel'` it does not work and will print warnings in the console (only on dev):

> The width being used (`400`) should be added to `image.screens`. Defaulting to `575`. Warning originated from `/img/my-image.png`.
> The width being used (`800`) should be added to `image.screens`. Defaulting to `991`. Warning originated from `/img/my-image.png`.
> The width being used (`600`) should be added to `image.screens`. Defaulting to `767`. Warning originated from `/img/my-image.png`.

What I dont understand is why the vercel-provider is even making a check like this. `image.screens` not related to the image width, it is used for defining screen sizes and breakpoints to load different image widths. It makes no sense to restrict image resizing just to these screen size breakpoints.

So I created a PR to remove this restriction: #1181

MickL:
For some reason it was not possible to just use any width, instead all width's that want to be used needed to be predefined within `image.screens` which is related to screen size but not to image width. I removed this restriction.

Fixes #1180

### DIFF
--- a/src/runtime/providers/vercel.ts
+++ b/src/runtime/providers/vercel.ts
@@ -4,7 +4,6 @@ import type { ProviderGetImage } from '../../types'
 // https://vercel.com/docs/more/adding-your-framework#images
 
 export const getImage: ProviderGetImage = (src, { modifiers, baseURL = '/_vercel/image' } = {}, ctx) => {
-  const validWidths = Object.values(ctx.options.screens || {}).sort((a, b) => a - b)
   const largestWidth = validWidths[validWidths.length - 1]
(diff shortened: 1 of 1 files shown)


danielroe:
See https://github.com/nuxt/image/issues/1180#issuecomment-1879198450.

PR #1224, opened 2024-01-30
fix: use ipx for static builds instead of auto detect

pi0:
Context: We discovered it with @Atinux investigating broken nuxt UI docs since it was using vertical image optimization and not picked from.. Another story probably to solve for another day...

---

When building Nuxt in static mode (`nuxt generate`) we assume everything is at build-time. Auto detected providers (vercel, amplify) only benefit for production deploys with a server running where as for a static deploy, we can save (dynamic optimization) costs and reduce issue surface by optimizing at build time.

This PR improves `detectProvider` internal utility to preserve auto-detected status and later use in nitro hook to switch back to ipx-static if needed (so user input is still respected)

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -131,10 +131,10 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
     })
 
     nuxt.hook('nitro:init', async (nitro) => {
-      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic' || _detectedProvider.auto) {
+      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic' || _detectedProvider?.auto) {
         const resolvedProvider = nitro.options.static || options.provider === 'ipxStatic'
           ? 'ipxStatic'
-          : nitro.options.node ? 'ipx' : 'none'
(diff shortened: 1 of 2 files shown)


atinux (approved):

danielroe (dismissed):

danielroe:
Actually, thinking about this I'm not 100% sure this is the right choice. Even with statically generated builds, we don't know that all images will be optimised at build time. (For example, they may be loading them in from an API.)

I think it's safer to have users opt-in to this behaviour.

pi0:
Vercel default config does allow external URLs by default and we can detect this at build time using config if we want to allow this also automatically.

Same goes for Netlify, our second partner that is pushing their CDN optimization

Build-time cost, is always less than on demand and we can use build-time caching and for staticly build-websites by default, we optimize only build-time images (again unless configured to allow external domain)

So I guess question is that Nuxt is aiming to make providers happy/reach or users.

danielroe:
The point is that we can optimise all images that are _known_ at generate time, and for plenty of websites that is perfect 👌 

But for websites that have any kind of dynamic image which might rely on the current behaviour, this would be a breaking change. That's why I'm suggesting it should be opt-in.

PR #1231, opened 2024-02-01
fix: command palette bug

Flosciante:


danielroe:
It feels like the lock file change here is too big. Would you have a look?

Flosciante:
It's weird, i take a look as soon as possible sorry for that

PR #1234, opened 2024-02-03
feat: support Netlify Image CDN

ascorbic:
In September lase year, Netlify deprecated Large Media, which is currently the basis of the `netlify` image provider. The DX was always awkward and it was never widely used. Subsequently we launched [Netlify Image CDN](https://docs.netlify.com/image-cdn/overview/), which is a new product that is available automatcially on all sites, and is much easier to use. This PR adds support for the new image CDN, and makes it the default for the `netlify` provider. To avoid breaking existing sites that use Large Media, I have done this by renaming the old provider to `netlifyLargeMedia`, creating a new provider called `netlifyImageCdn`, and then replacing the `netlify` provider with one that conditionally exports the other provider, depending on whether Large Media is detected on the site. Users can also choose one of the providers manually.

This is a draft to get feedback on my approach. If it looks good I'll update the tests and docs too. Because the image CDN is automatically enabled on all sites, I would like to follow up with a PR to auto-enable this provider, similarly to the Vercel provider.


src/runtime/providers/netlify.ts
```diff
@@ -1,50 +1,9 @@
...
-      fill: 'smartcrop',
-      contain: 'fit'
-    }
-  },
-  joinWith: '&',
-  formatter: (key, value) => `${key}=${value}`
-})
+// If the site has the deprecated Netlify Large Media enabled we will use the old
+// Large Media adapter. Otherwise, we will use the Netlify Image CDN
+const hasLargeMedia = Boolean(process.env.NETLIFY_LFS_ORIGIN_URL)
```
danielroe, line 6: If `NETLIFY_LFS_ORIGIN_URL` is available in the build-time environment it would be worth doing this in the build/module scope rather than in the runtime. That way we avoid bundling both implementations.

ascorbic:
@danielroe OK, I've changed it to now detect at build time instead.

src/provider.ts
```diff
@@ -136,7 +142,14 @@ export async function resolveProvider (_nuxt: any, key: string, input: InputProv
...
-  aws_amplify: 'awsAmplify'
+  aws_amplify: 'awsAmplify',
+  netlify: 'netlify'
+}
+
+const normalizableProviders: Partial<Record<string, () => ImageProviderName>> = {
+  netlify: () => {
+    return process.env.NETLIFY_LFS_ORIGIN_URL ? 'netlifyLargeMedia' : 'netlifyImageCdn'
+  }
 }
```
ascorbic, line 153: I've added this as a new way for providers to resolve at build time. It alows an alias or detected platform to provide logic to resolve to one or other defined provider types.

src/runtime/providers/netlifyImageCdn.ts
```diff
@@ -0,0 +1,53 @@
...
+    position: {
+      top: 'top',
+      right: 'right',
+      bottom: 'bottom',
+      left: 'left',
+      center: 'center'
+    }
+  },
+  joinWith: '&',
+  formatter: (key, value) => encodeQueryItem(key, value)
```
danielroe, line 37: @ascorbic Just updated this but want to confirm - is it correct that we should encode query items, e.g. `?url=%2Fimage.jpg` instead of `?url=/image.jpg` or do you prefer to keep the slashes unencoded?

ascorbic: @danielroe It looks like either works, but I reckon encoding is safer

src/provider.ts
danielroe: Seems good to me, and it's internal so we can always refactor in future.

src/provider.ts
```diff
@@ -136,7 +142,14 @@ export async function resolveProvider (_nuxt: any, key: string, input: InputProv
 
 const autodetectableProviders: Partial<Record<ProviderName, ImageProviderName>> = {
   vercel: 'vercel',
-  aws_amplify: 'awsAmplify'
+  aws_amplify: 'awsAmplify',
+  netlify: 'netlify'
```
danielroe, line 146: Adding a comment that we may change the logic for these providers in a subsequent PR. I'm discussing with @pi0 whether it's a breaking change to enable runtime support when users are performing fully static builds (e.g. `nuxt generate`).

context: https://github.com/nuxt/image/pull/1224

danielroe (approved):

danielroe:
Thank you so much ❤️

Anoesj:
Very nice change, but it would have been better if this had been marked as a breaking change. Netlify deployments now seem to use `netlify` as a provider by default, which is a change and broke all images in an app of mine hosted on Netlify after deployment, because I was assuming IPX to be used by default at all times. Changing provider to `ipx` manually in `nuxt.config.ts` fixed the issue.

Context: my app is built using `NITRO_PRESET=netlify-static nuxt build` with `ssr: false` in `nuxt.config.ts` to generate a preview of the app as a plain Vue app. The images are all "passthrough", so there is no image optimization going on at all, not by IPX either.

PR #1248, opened 2024-02-14
feat: add `useBackgroundImage` composable

shunnNet:
Add a `useBackgroundImage` composable to enable the background to use the same functionality as `<nuxt-img>`.

## Usage
```vue
<script setup lang="ts">
  const cls = useBackgroundImage('/images/nuxt.png', {
    provider: 'ipx',
    sizes: '200,500:500,900:900',
    modifiers: {
      width: 300,
      height: 300,
      // others...
    },
    densities: "1x 2x",
    preload: true,
    nonce: 'any nonce',
  })
</script>
<template>
  <div
    :class="cls"
    :style="{
      width: '300px',
      height: '300px',
      backgroundSize: 'cover',
      backgroundRepeat: 'no-repeat',
      backgroundPosition: 'center'
    }"
  />
</template>
```

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -102,6 +102,11 @@ export default defineNuxtModule<ModuleOptions>({
       from: resolver.resolve('runtime/composables')
     })
 
+    addImports({
+      name: 'useBackgroundImage',
+      from: resolver.resolve('runtime/composables')
+    })
+
     // Add components
     addComponent({
       name: 'NuxtImg',
--- a/src/runtime/composables.ts
+++ b/src/runtime/composables.ts
@@ -1,9 +1,9 @@
-import type { $Img } from '../types'
-
+import type { $Img, ImageSizesOptions } from '../types'
+import { generateRandomString } from './utils'
 import { createImage } from './image'
 // @ts-expect-error virtual file
 import { imageOptions } from '#build/image-options'
-import { useNuxtApp, useRuntimeConfig } from '#imports'
+import { useNuxtApp, useRuntimeConfig, useHead } from '#imports'
 
 export const useImage = (): $Img => {
   const config = useRuntimeConfig()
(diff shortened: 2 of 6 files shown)


shunnNet:
Hello, I have added 3 new commits:

- During SSR, because class names are randomly generated, it leads to duplicated insertion of style tags on both the server and client sides. Therefore, I attempted to fix this issue. Commit: [bf2665d0](https://github.com/nuxt/image/pull/1248/commits/bf2665d079ac4272d8f6f452c287e96e36bbecd1)
- I initially thought there might be a similar duplication issue with preload, but it seems to occur only in dev mode. Therefore, I added a condition to check for this and removed that condition later.

danielroe:
My apologies for the slow review on this PR! I'm personally not sure this would be the right way to add this functionality. (Adding CSS generated in JS.)

Instead I think generating the image URL with `useImage()` and passing it in to a `<style>` block [with a binding](https://vuejs.org/api/sfc-css-features.html#v-bind-in-css) might be worth trying.

shunnNet:
First of all, thank you for your response.

I believe I should first explain the rationale behind this PR:

nuxt-img encapsulates the functionalities of html `` and `<picture>` into a simple interface, so intuitively, I thought it should have a similar solution for `background-image` as well. However, the fact is that the documentation only suggests users manually generate `src` and place them either in `v-bind:style` or in a `v-bind` within a style block. If users want functionality similar to the `<NuxtImg>` component, they have to write scripts themselves to generate different versions of the image and set CSS accordingly.

Therefore, the purpose of this PR is to make setting `background-image` as convenient as `NuxtImg`. I believe it would make the package more complete. I think `v-bind` cannot address this situation.

However, you mentioned uncertainty about whether "Adding CSS generated in JS." is a good approach. I would like to understand more about your concerns regarding this idea. Could you please elaborate on why this might not be the correct approach and what your concerns are?

PR #1254, opened 2024-02-16
fix(uploadcare): omit base when resolved URL is returned

aaronLejeune:
**Problem:**
There is no option to disable the URL prefix in the UploadCare settings.

It would be nice to have an option in the nuxt.config.js where if cndURL is false, there is no URL prefix.


danielroe:
If you pass `/` as a URL, does it behave as you wish?

aaronLejeune:
Sadly, not.. the generated url will look like this:
`<img src="/https://ucarecdn.com/5d...53/-/format/auto/-/quality/smart/-/resize/2560x/"`
with this config
`image: {
    uploadcare: {
      cdnURL: '/'
    }
  },`
I've found no workaround aside from editing the code :)

aaronLejeune:
Please let me know if there is anything I can do :)

danielroe:
Closing as I can't reproduce the original bug and I believe this feature is already implemented, but please do feel free to provide a reproduction of the original issue and we can definitely reopen.

aaronLejeune:
Sorry for the late response!
Here is a reproduction of the issue:

Screenshot 2024-04-05 at 11 15 41

https://stackblitz.com/edit/github-kzrcuy-6sn53l?file=app.vue,server%2Fapi%2FheadlessCMS.ts,server%2Ftsconfig.json

aaronLejeune:
Hey @danielroe , feel free to let me know if I need to clarify a bit more ;)

danielroe:
Oh, I see - if the returned data from CDN already has a protocol. I think the fix instead should be simply not to prefix with `cdNURL` in that case.

We can test with utilities provided from `ufo`.

danielroe (approved):

Issue #1264, opened 2024-02-26
Providers not being built into nuxt output


jfelicianiats:
### Issue
When running a production build of the application where the `node_modules` have been pruned via `npm prune --production`, if you don't include `ipx` in the "dependencies" section of your package.json, the application will fail to load images and if you check the network tab, the following error is listed for the image resources.  Even installing `ipx` as a `devDepdendency` fails in the final build if node_modules are not present.  Im assuming the ipx lib is not being built into the final nuxt output.

```
[500] [IPX_ERROR] Could not load the "sharp" module using the linuxmusl-x64 runtime
Possible solutions:
- Ensure optional dependencies can be installed:
    npm install --include=optional sharp
    yarn add sharp --ignore-engines
- Ensure your package manager supports multi-platform installation:
    See https://sharp.pixelplumbing.com/install#cross-platform
- Add platform-specific dependencies:
    npm install --os=linux --libc=musl --cpu=x64 sharp
- Consult the installation documentation:
    See https://sharp.pixelplumbing.com/install
```

### Current Workaround
Installing the `ipx` library as a `dependency` and **NOT** a `devDependency` seems to remedy the issue.

### Not working package.json
```json
{
  "name": "nuxt-app",
  "private": true,
  "type": "module",
  "scripts": {
    "build": "nuxt build",
    "dev": "nuxt dev",
    "generate": "nuxt generate",
    "preview": "nuxt preview",
    "postinstall": "nuxt prepare"
  },
  "devDependencies": {
    "@nuxt/image": "^1.3.0",
    "@nuxtjs/eslint-config-typescript": "^12.1.0",
    "@nuxtjs/eslint-module": "^4.1.0",
    "@nuxtjs/i18n": "^8.0.0",
    "@nuxtjs/tailwindcss": "^6.11.0",
    "@types/lodash-es": "^4.17.12",
    "@typescript-eslint/parser": "^6.19.1",
    "@vee-validate/yup": "^4.12.5",
    "@vueuse/core": "^10.8.0",
    "eslint": "^8.56.0",
    "eslint-config-prettier": "^9.1.0",
    "eslint-plugin-prettier": "^5.1.3",
    "nuxt": "^3.9.3",
    "prettier": "^3.2.4",
    "rimraf": "^5.0.5",
    "sass": "^1.70.0",
    "typescript": "^5.3.3",
    "vee-validate": "^4.12.5",
    "vue": "^3.4.14",
    "vue-router": "^4.2.5",
[...]

Issue #1262, opened 2024-02-26
Just don't work


wireinet:
I just can't make this work...
install, add module, restart server, get error:
[worker reload] [worker init] Cannot find package 'ipx' imported from D:\......\.nuxt\dev\index.mjs
just default setup: no option in 'defineNuxtConfig'

gbyesiltas:
I want to note something here:

Even after the PR that @danielroe linked, I continued to have this issue. Then when I dove a bit deeper to create a reproduction I realised that the difference between the working version and my non-working version was the way that the `ipx` package was imported in the `.nuxt/dev/index.mjs` file. In the working version, the import path went like `"file:///Users/..."` where as the non-working version it was simply `"ipx"`. 

Trying to figure out where this was coming from, I did a bunch of things: (1) Removing the explicit `@nuxt/devtools` dependency from our `package.json` since it now automatically comes from nuxt (2) Copying over the `excludes` field from the auto generated nuxt `tsconfig` again to the project one since we overwrite/expand it, and (3) just went back-and-forth between versions 1.3.0, 1.4.0, and 1.6.0. By the end of this, the issue was fixed for me. 

I'm not sure which one of them ended up fixing it but maybe this will be helpful to someone 🤷

Issue #1257, opened 2024-02-20
`sharp` optional binaries not included in `.output`


Aareksio:
### Reproduction
```
git clone git@github.com:nuxt/image.git
cd image
pnpm install
pnpm dev:prepare
pnpm build
pnpm dev:build
rm -rf node_modules/@img
node playground/.output/server/index.mjs
```
Open: http://localhost:3000/provider/ipx

---

```
[500] [IPX_ERROR] Could not load the "sharp" module using the ... runtime
```

---

Sharp binaries, bundled as `@img/sharp-*`, are not included in generated `.output`. Tests may pass, because of module resolution. We need to isolate `.output` (or delete root project `node_modules`, as with reproduction), as it _should be done_ in production.

Related: https://github.com/lovell/sharp/issues/4000 (cross architecture build)
Related: #1237, #1210 (either cross architecture build or isolating .output)
Related: #1253 (both cross architecture build and isolating .output)

---

Workaround before this gets fixed: downgrade `sharp` to `0.32.6` via your package manager.

tasiotas:
I couldn't get it working with `pnpm` and `supportedArchitectures`. Had to go back to Yarn... 😟

alimozdemir:
This is a blocker for me right now, either I will remove nuxt-image or I will find a workaround with pnpm. It is very annoying

juane1000:
The latest update (1.4.0) shoots my server build to 50MB+ from ~12-ishh. Is this intentional/desirable? Seem exorbitant. I was unaware that the the plugin included resizing functionality. Is there an option to exclude the "sharp" dependency from the server bundle and have it installed in the deployment server environment? Thanks.

oripka:
> Has anyone found a workaround with pnpm? This is also a blocker for me at the moment.
> 
> UPDATE: added sharp 0.33.3 to overrides and it works correctly.

This is the only fix that work for me, everything else described even using the latest nuxt image version failed when deploying a nuxt project built on m1 to a linux x64 server.

oripka:
I just moved to a monorepo setup and had to include `nuxt-image`  and the override in the root package.json in order to make this work. First I had `nuxt-image` as a dependency in the base layer and no matter where I applied overrides it would not work.

```
{
    "devDependencies": {
        "@nuxt/image": "^1.7.0"
    },
    "pnpm": {
        "overrides": {
            "sharp": "0.33.4"
        },
        "supportedArchitectures": {
            "os": [
                "current",
                "linux"
[...]

Issue #1253, opened 2024-02-16
[500] [IPX_ERROR] Could not load the \"sharp\" module using the linuxmusl-x64 runtime


vitaliytsoy:
![image](https://github.com/nuxt/image/assets/22080639/2b7add09-0382-4bf3-b0fe-495cf4046b0b)

Have a problem with the sharp module required by IPX dependency. Some images return 500 status with error.
Happening because of the different installation and runtime environments.

Installation(`pnpm i`) and build(`pnpm build`) runs on ubuntu machine in github action. 
Then production build copied to Docker container that runs on **node:20-alpine** linux which uses musl libc and require `sharp` to use different binaries - specific for alpine environment (`linuxmusl-x64`).

Tried to install sharp manually as onptional dependency with `pnpm.supportedArchitectures`([link](https://pnpm.io/package_json#pnpmsupportedarchitectures)), had no effect - target binaries are still not present in production build. 
```
  "pnpm": {
    "supportedArchitectures": {
      "os": ["linux", "current"],
      "cpu": ["x64", "current"],
      "libc": ["musl", "current"]
    }
  },
```

Found pretty ugly fix, which requires you to reinstall `sharp` package in server production build.
Run `npm install sharp --force --omit=dev --cpu=x64 --os=linux --libc=musl`
Inside `.output/server` folder
Which basically installs `@img/sharp-linuxmusl-x64` - sharp's optional dependency inside server production build. 

Is it possible to take in consideration `supportedArchitectures` when building production build and add required dependency in server's node_modules?

Setup:

pnpm: v8.15.2
node: v20.7.0
@nuxt/image: ^1.3.0
nuxt": ^3.8.2

slantz:
I had similar problem recently:
- multi-stage dockerfile
- `pnpm@8.9.0`
- docker image `node:20.8.0-alpine`
- `nuxt@3.10.1`

At the end after checking all the installed dependencies under the `.output/server/node_modules` the `sharp` distribution for linux was there, since we have it pinned in the `pnpm-lock.yaml` the project was working but always missing images.

But the problem was in how `pnpm` resolves the dependencies. Unlike `npm` or `yarn`, `pnpm` uses a content-addressable filesystem to store its packages, which results in a more efficient disk space usage and faster installations. However, this approach means that `pnpm` heavily relies on symlinks within the `node_modules` directory to link back to its global store, which is typically located in `node_modules/.pnpm`.

[...]

PR #1265, opened 2024-02-26
fix(ipx): temporarily downgrade ipx/sharp

Aareksio:
Downgrade ipx to 2.1.0 to avoid issues with `sharp` (#1257, #1210, #1237, #1253, #1262, #1264, #1228, #1227) until https://github.com/unjs/ipx/issues/215 is resolved. [@pi0 comment on Discord](https://discord.com/channels/473401852243869706/847861019421179994/1211757315070627910).

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "eslint": "8.57.0",
     "globby": "^14.0.1",
     "happy-dom": "^13.5.0",
-    "ipx": "^3.0.1",
+    "ipx": "^2.1.0",
     "jiti": "1.21.0",
     "nuxt": "^3.10.3",
     "playwright-core": "^1.41.2",
@@ -63,7 +63,7 @@
     "vue-tsc": "^1.8.27"
   },
   "optionalDependencies": {
-    "ipx": "^3.0.1"
+    "ipx": "^2.1.0"
   },
   "packageManager": "pnpm@8.15.3",
   "resolutions": {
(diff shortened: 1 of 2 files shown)


alimozdemir:
Any news? @pi0

pi0 (approved):

pi0:
(agree as hotfix, although, as asked before, with all shared linked, i expected to have a **minimal reproduction** with clear steps and ideally only nuxt+ipx or nitro+ipx. All linked issues lack it, so whatever we do, we burry the root cause and replace it with WASM.)

Aareksio:
Downgrading was never meant to be a long term fix. The fix is to `nuxt-image`, therefore the minimal reproduction is `nuxt` + `nuxt-image`. With clear steps. Run the official example. It is broken. #1257

Nuxt is in the top 5 most popular web frameworks and yet one of the most popular, **official** module is reported as broken for 1.5+ months. If there is no resolution to such critical issue within such timeframe, slapping the band aid, given how simple it is, seems to be a sensible short-term solution. Then we have all the time in the world to figure the root issue without hundreds of users and companies suffering.

As primarily business person, I am sometimes a bit salty about the balance between _fun_ and _maintenance_ in nuxt ecosystem. Conference season probably does not help 😅 I'll try to prepare plain nito + ipx reproduction tonight as requested, so you can track the exact cause. Esbuild shenanigans are beyond my comparison _yet_ :(

pi0:
You can try on edge. I let @danielroe handle the release for the meanwhile as hotfix.

alimozdemir:
Hi @danielroe What are the plans for the next release?

PR #1268, opened 2024-03-01
feat: encode reserved characters in URI path

riddla:
I fell into a rabbit hole yesterday analysing why Slack would not show the preview of an `og:image`. After a long try and error  I discovered that the URIs generated via the `ipx` provider contains ampersands respectively: the Slackbot needs URIs with encoded reserved characters in the URIs path .

With encoded characters (`fit_inside%26f_png%26s_1200x630`) the preview came to life.

### DIFF
--- a/src/runtime/providers/ipx.ts
+++ b/src/runtime/providers/ipx.ts
@@ -23,7 +23,9 @@ export const getImage: ProviderGetImage = (src, { modifiers = {}, baseURL } = {}
     delete modifiers.height
   }
 
-  const params = operationsGenerator(modifiers) || '_'
+  let params = operationsGenerator(modifiers) || '_'
+
+  params = encodeURIComponent(params)
 
   if (!baseURL) {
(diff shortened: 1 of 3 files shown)


danielroe:
I'm opening a new PR as I don't have permission to push to this one: https://github.com/nuxt/image/pull/2005

danielroe:
@riddla feel free to pull in my commits in the linked branch on nuxt/image into this PR.

testing it out in the playground, it doesn't seem to work. I get this error:

```
[500] [IPX_ERROR] Expected number between 0.3 and 1000 for sigma but received undefined of type undefined
```

riddla:
@danielroe, the change in [#08fcb23](https://github.com/nuxt/image/pull/1268/commits/08fcb236d13dadd149e5c67e66d49190dab54025) sorts  the parameters in alphabetical order.

Which did fix the ipx image not showing up in the playground. But to be honest, I did not fully grasp the underlying issue in ipx, so this pragmatic approach might not be sufficient.

PR #1287, opened 2024-03-15
feat: support domains option with Netlify Image CDN

ascorbic:
Adds support for [configuring remote images](https://docs.netlify.com/image-cdn/overview/#remote-path) in Netlify Image CDN. Previously this required manual configuration in the `netlify.toml`. With this PR, the provider sets the value automatically based on the `image.domains` option.

Depends on https://github.com/unjs/nitro/pull/2264


danielroe (approved):
Looks good to me. Just waiting for Nitro PR to merge.

danielroe:
Fantastic! I’ll just confirm that domains are always normalised to without protocol and then we should be good to go…

Issue #1299, opened 2024-03-23
See error if src is undefined.


cfdxkk:
### Veersion
``` yaml
dependencies:
  '@nuxt/image':
    specifier: ^1.4.0
    version: 1.4.0
  nuxt:
    specifier: ^3.10.3
    version: 3.10.3(vite@5.1.5)
  vue:
    specifier: ^3.4.19
    version: 3.4.21
  vue-router:
    specifier: ^4.3.0
    version: 4.3.0(vue@3.4.21)
```

### Issue
ERROR: `input must be a string (received undefined: undefined)`

In my VS Code, It shows that src is optional and can be string or undefined , but if I actually use undefined I get this error.

My TypeScript was broked?

![image](https://github.com/nuxt/image/assets/41364445/e44dd5b0-e9de-420d-876b-bf9a3263d568)
![image](https://github.com/nuxt/image/assets/41364445/49077f56-b95f-4183-bf68-6abee7a6ec41)

### Minimal Reproduction
Source Code: https://github.com/cfdxkk/nuxt-image-undefined-src

``` js
app.vue

<template>
	<NuxtImg :src="undefined" />
</template>
```
or
``` js
app.vue

<script setup lang="ts">
	const imageUrl = ref<string>();
	if (false) {
		imageUrl.value = 'foo.jpg'
	}
</script>

<template>
	<NuxtImg :src="imageUrl" />
</template>
```

cfdxkk:
update:

Use `v-if` with ':src' can temporary fix.
``` vue
<script setup lang="ts">
	const imageUrl = ref<string>("");
</script>
<template>
	<NuxtImg v-if="imageUrl" :src="imageUrl" />
</template>
```

oritwoen:
I had a similar problem too. Quite by accident, I discovered that adding the parameter: `placeholder` with the correct url link will display this placeholder image instead of an error if `src` is empty or has values not defined correctly. It's not a perfect solution but it works.

danielroe:
resolved in https://github.com/nuxt/image/pull/1300.

PR #1300, opened 2024-03-24
fix: handle `undefined` or empty image source values

ThimoDEV:
References #1299 

I have been thinking about the best solution for this issue. Typescript says by design that the src prop is string | undefined so we either need to narrow that down or allow it?

Vue by design always adds undefined type as far as I could see. When I removed the throw error it works more as expected and with the reuired property on the src prop I get a Vue warning that it is not set which is nice.

However when I set:

`:src="undefined"` I get a Vue warning the the type for the src prop is incorrect. 

Is it better to match the typescript types (and it quite follows the normal  behaviour or have it generate a Vue prop required warning with the above downside?


toyi:
Allowing `undefined` would also allow the developer to eventually catch this `undefined` `src` in a custom provider and do something else with it (showing a specific placeholder or something else).

Currently, we must pass a string (and `''` doesn't work, it throws an error saying it received `''` instead of a string). So when the image _can_ be `undefined`, it must be handled directly where you use the component (or in a high order component).

danielroe (approved):

PR #1310, opened 2024-04-03
docs: use new `nuxi module add` command in installation

danielroe:
This updates the documentation to use the [`nuxi module add` command](https://github.com/nuxt/cli/pull/197) which should simplify docs a bit and also improve user experience as there's no need to add to `nuxt.config` manually.

It's documented [here](https://nuxt.com/docs/api/commands/module#nuxi-module-add).

I may have missed a few spots in the documentation as I'm doing this across the modules ecosystem assisted by the power of regular expressions ✨, so I'd appreciate a review 🙏

### DIFF
--- a/docs/pages/index.vue
+++ b/docs/pages/index.vue
@@ -13,7 +13,7 @@ useSeoMeta({
   ogImage: 'https://image.nuxt.com/social-card.png',
   twitterImage: 'https://image.nuxt.com/social-card.png',
 })
-const source = ref('npm i @nuxt/image')
+const source = ref('npx nuxi@latest module add image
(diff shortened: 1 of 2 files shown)


docs/pages/index.vue
```diff
@@ -13,7 +13,7 @@ useSeoMeta({
   ogImage: 'https://image.nuxt.com/social-card.png',
   twitterImage: 'https://image.nuxt.com/social-card.png',
 })
-const source = ref('npm i @nuxt/image')
+const source = ref('npx nuxi@latest module add image
```
danielroe, line 16:
```suggestion
const source = ref('npx nuxi@latest module add image')
```

Issue #1317, opened 2024-04-10
Lazy loading still loads all images after first page load


Ragura:
Reproduction: https://stackblitz.com/edit/github-roaj2c?file=pages%2Findex.vue

I've noticed that unlike native `<img>` elements, `<NuxtImg>` doesn't seem to handle `loading="lazy"` correctly after the first time a page is loaded. Instead, all images are loaded immediately.

In the linked reproduciton you can see this in action. Keep an eye on the network tab.
The app starts on `index.vue`. If you navigate to the NuxtImg page using the provided link, only the images in the viewport will be loaded as they should. Now navigate back to the index page and back to the NuxtImg page. You will see that all images are loaded immediately, including those far below the viewport.

I have also provided a page that uses regular `<img>` tags. You will notice that those do not have the same issue and act as they should even after page navigation.

I hope I'm not misunderstanding what `<NuxtImg>` does with lazy loading. I assume it does the same as a native `<img>` tag and it should work accordingly.

danielroe:
The behaviour here seems to be exactly the same between the two pages. In my browser it loads six images, and then lazy loads the rest.

http://volta.s3.fr-par.scw.cloud/Clean_Shot_2024_04_22_at_15_07_25_b7045710a5.mp4

Ragura:
That's not the behaviour I'm getting. I followed the exact same steps as you did in the video and uploaded my own version here:
https://drive.google.com/open?id=1DdjOHel5dUGazqKcBHuWl9W5_KddZb6N
Using Arc as my browser (but I think you are too?).

danielroe:
Looking at this, it seems to be a Vue upstream issue. See https://stackblitz.com/edit/github-hec3gw?file=src/App.vue.

Would you raise an issue in https://github.com/vuejs/core, perhaps? 🙏

Ragura:
Thanks for looking into it. I've reraised the issue in the VueJS core repo here: https://github.com/vuejs/core/issues/10764

PR #1338, opened 2024-04-23
fix: move src attribute to end of <img> tag to fix lazy loading (#1317)

Ragura:
As described in this Vue core issue (https://github.com/vuejs/core/issues/10764), using the `src` attribute before `loading` results in the lazy-loading being inconsistent across loads. This PR moves the `src` tag to the last position in the element.

I tested the behavior before and after the change in the playground to confirm it is working correctly. If you want I can also commit the playground page for reference.

Feel free to correct me on anything I might have done wrong, be it in the code or while making this PR. I'm not very experienced with the process of contributing to public repos yet!

### DIFF
--- a/src/runtime/components/nuxt-img.ts
+++ b/src/runtime/components/nuxt-img.ts
@@ -147,11 +147,11 @@ export default defineComponent({
 
     return () => h('img', {
       ref: imgEl,
-      src: src.value,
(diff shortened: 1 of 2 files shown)


danielroe (approved):
Perfect! 👌

PR #1361, opened 2024-05-10
feat(cloudinary): support reading project from absolute url

markbrouch:
Resolves https://github.com/nuxt/image/issues/1376

Adds the ability for a full cloudinary url to be used as `src` instead of relying on `baseURL` to construct the image url.

For example:
```vue
<NuxtImg src="https://res.cloudinary.com/example/image/upload/example.jpg" />
```
should use `https://res.cloudinary.com/example/image/upload` as the `baseURL`, regardless of what `baseURL` is configured as.

This is needed for apps that receive a full cloudinary url from the backend which may be responsible for determining the specific cloud and base url.

See discussion in Discord thread: https://discord.com/channels/473401852243869706/1238298456540577802


danielroe:
Would you be able to add an example to the test suite? 🙏

markbrouch:
@danielroe - I have added an example to the test suite in https://github.com/nuxt/image/pull/1361/commits/cc869978c6523a73333c42f1904982a1833e5ab6

danielroe (approved):
Thank you ❤️

PR #1415, opened 2024-07-23
fix(hygraph): new asset management support

timbenniks:
- Cleaner URL splitter for new Hygraph asset management.
- Added image_auto option in playground example
- Updated snapshots


src/runtime/providers/hygraph.ts
```diff
@@ -57,7 +79,7 @@ export const getImage: ProviderGetImage = (
   } = modifiers
 
   if (!baseURL) {
-    throw new Error('No Hygraph image base URL provided.')
+    throw ('No Hygraph image base URL provided.')
```
danielroe, line 82:
```suggestion
    throw new Error('No Hygraph image base URL provided.')
```

danielroe (approved):

PR #1420, opened 2024-08-05
fix(setup): Setup provider using the `provider` key

### LINKED ISSUE #1414: IPX provider not working properly when the default provider is not IPX
I have custom provider (thumbor) as my default provider in the config and in some case I still need IPX provider, so I pass provider props in the `NuxtImg` components as `<NuxtImg provider="ipx" />` and the result is 404. `_ipx` path is not found.

The config I had:
```ts
image: {
  ipx: {},
  provider: "thumbor",
  providers: {
    thumbor: { ... }
  }
}
```

When I omit the `ipx` property, it throw `Unknown provider: IPX` error. The config works when I replace the default provider with `ipx`.

Is it expected behavior?

I create a minimal reproduction using `ipx` and `unsplash` provider to suit my problem: https://stackblitz.com/edit/nuxt-image-ipx-repro?file=app.vue

Thank you!

Tofandel:
### 🔗 Linked issue

Fixes #1414

### ❓ Type of change

### 📚 Description

Because the name is user provided and can (and should be different if we have multiple instances of one provider) then any name different than the provider name (eg: `my-remote-ipx` instead of `ipx`) will not setup the provider as expected

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -137,11 +137,12 @@ export async function resolveProvider(_nuxt: any, key: string, input: InputProvi
   }
 
   const resolver = createResolver(import.meta.url)
+  
+  const setup = input.setup || providerSetup[input.provider as ImageProviderName] || providerSetup[input.name as ImageProviderName]
+  
   input.provider = BuiltInProviders.includes(input.provider as ImageProviderName)
     ? await resolver.resolve('./runtime/providers/' + input.provider)
     : await resolvePath(input.provider)
-
-  const setup = input.setup || providerSetup[input.name as ImageProviderName]
 
   return <ImageModuleProvider> {
     ...input,


src/provider.ts
```diff
@@ -137,12 +137,13 @@ export async function resolveProvider(_nuxt: any, key: string, input: InputProvi
   }
 
   const resolver = createResolver(import.meta.url)
+  
+  const setup = input.setup || providerSetup[input.provider as ImageProviderName] || providerSetup[input.name as ImageProviderName]
+
```
Tofandel, line 142:
```suggestion

```

Tofandel:
Drafting this for now as there is some weird things going on with ipx

```
if (typeof p.setup === "function" && p.name !== "ipx" && p.name !== "ipxStatic") {
        await p.setup(p, options, nuxt);
}
```

Because we can't use nitro so early

danielroe:
would you be able to add an example of the usage you have in mind, perhaps by adding it to the test fixture? 🙏

Issue #1329, opened 2024-04-17
Images cannot be parsed when obtained through event handlers


GalacticHypernova:
Currently NuxtImg only seems to work fine with images that are presented at build time in `~/public`. While that may suffice most static sites, for sites that require user interaction, including uploading media, this ends up breaking the images in production. It is not ideal to have to rebuild the app every time an image is uploaded, for quite obvious reasons.
Take for example this code:
```ts

// ~/server/routes/images/[image].get.ts
export default defineEventHandler({
    handler: async (event) => {
        const img = getRouterParam(event, "image")
        const imageStorage = useStorage("images") // this could, for example, be an FS driver configured in `nuxt.config.ts`
        if (await imageStorage.hasItem(img)) {
            return await imageStorage.getItem(img)
        }
    }
})
```
This works fine when going to it on the browser, but NuxtImg fails to parse this, with a message stating the file cannot be found.

It would be very beneficial to have the benefits of image optimization for images that are obtained dynamically through event handlers, as most SSR apps use this kind of functionality.

GalacticHypernova:
Hey @Baroshem ! I added the reproduction of the event handler that NuxtImg can't pick up. Daniel also stated that it is currently a limitation of NuxtImg, that it can only reach images present in `~/public` (or those from external services). Is there any part of the reproduction that you want specifically or would you like just a full minimal project to demonstrate it? Because the event handler I sent here can be real easily attached to a project and then referenced in the src of the image, which will lead to the error.

GalacticHypernova:
Hey again @Baroshem !

I made some more digging, and I think I understand why this is happening.

As long as you don't outright disable FS storage, Nuxt automatically registers IPX FS:
https://github.com/nuxt/image/blob/27dbb2e36edbb11ce11cc606a09ca3ca92d10e4e/src/ipx.ts#L46C5-L49C7


IPX, in response, uses the FS storage when you provide an image, regardless of whether it's actually a public asset or something fetched on the server-side:
https://github.com/unjs/ipx/blob/main/src/ipx.ts#L102C4-L104C48

So if you have a route named `/images/[image].get.ts` (this might even be happening with direct placement of the image file there, though I haven't tested), any call to it would get treated as an FS call to /public/images, thus failing to parse images from a server route.

Baroshem:
Thanks @GalacticHypernova for the additional digging and concept!

[...]

PR #1473, opened 2024-09-05
feat: add support for image helpers in nitro endpoints

### LINKED ISSUE #1447: add support for image helpers in nitro endpoints
I want to generate image url path on my server api response. it's my composable `useImageUrl.js`

```javascript
export default (src) => {
  const img = useImage()

  const imgUrl = img(`/${src}.png`, {
    format: 'webp',
    quality: 80,
  })

  return imgUrl
}
```

and my server api:

```javascript
import useImageUrl from '~/composables/utils/useImageUrl'


export default defineEventHandler(async (event) => {
  const { code } = event.context.params
  return {
     path: useImageUrl(code)
  }
})
```

but when I `yarn dev`, it shows the error:
```
[nuxt] [request error] [unhandled] [500] useImage is not defined
  at useImageUrl (./composables/utils/useImageUrl.js:2:15)
  at Object.handler (./server/api/deck/[code].get.js:65:1)
  at async ./node_modules/h3/dist/index.mjs:1975:19
  at async Object.callAsync (./node_modules/unctx/dist/index.mjs:72:16)
  at async toNodeHandle (./node_modules/h3/dist/index.mjs:2266:7)
  at async ufetch (./node_modules/unenv/runtime/fetch/index.mjs:9:17)
  at async $fetchRaw2 (./node_modules/ofetch/dist/shared/ofetch.37386b05.mjs:222:26)
  at async $fetchRaw2 (./node_modules/ofetch/dist/shared/ofetch.37386b05.mjs:263:14)
  at async $fetch2 (./node_modules/ofetch/dist/shared/ofetch.37386b05.mjs:268:15)
  at async ./pages/deck/[code].vue:69:18
```

Is it support on server side? or any other way let me generate image url?

thanks.

Afshar07:
Implement image helper utilities usage in nitro endpoints, by adding image options and nitro compatible utils in server directory.
But to be honest, it's my first contribution and i might (probably am) mess with a wrong config, so, it will be so kind for anyone to point out if i have done anything wrong!

### 🔗 Linked issue
resolves #1447 
resolves https://github.com/nuxt/image/issues/1329

### ❓ Type of change

### 📚 Description

Add an utility on server folder and auto import it into nitro, and also write image options into a virtual path in nitro.
Set a boolean to check if the utility is used in nitro environment, if it is, ignore prerendering static images.


src/runtime/image.ts
```diff
@@ -5,16 +5,15 @@ import { imageMeta } from './utils/meta'
...
   const ctx: ImageCTX = {
     options: globalOptions,
   }
 
   const getImage: $Img['getImage'] = (input: string, options = {}) => {
     const image = resolveImage(ctx, input, options)
-
     // Prerender static images
-    if (import.meta.server && import.meta.prerender) {
+    if (import.meta.server && import.meta.prerender && !isCalledInNitro) {
```
danielroe, line 16: why do we not want to hint to prerender static images if called from a server endpoint?

Afshar07: Thanks for the hint, i have tried the generate command and i see that beside this, i have a lot of unhandled situations!
I investigate the source code more precisely, review my code and come back with a solution ASAP.
Is there anything specific i need to know? like about writing tests or anything else?
Thanks for your time.

danielroe: you can add the virtual file with `addServerTemplate`

src/module.ts
```diff
@@ -180,3 +186,15 @@ function pick<O extends Record<any, any>, K extends keyof O>(obj: O, keys: K[]):
...
   return newobj
 }
+
+function generateImageOptions(providers: ImageModuleProvider[], imageOptions: Omit<CreateImageOptions, 'providers' | 'nuxt' | 'runtimeConfig'>): string {
+  return `
+  ${providers.map(p => `import * as ${p.importName} from '${p.runtime}'`).join('\n')}
+  
+  export const imageOptions = ${JSON.stringify(imageOptions, null, 2)}
+  
+  imageOptions.providers = {
```
pi0, line 196: I _think_ this causes tree-shaker to not work properly with assignments (so code will be in bundle even if util not used but again just guess).

src/runtime/composables.ts
```diff
@@ -1,15 +1,17 @@
...
 import { useNuxtApp, useRuntimeConfig } from '#imports'
 
-export const useImage = (): $Img => {
+export const useImage = (event?: H3Event): $Img => {
   const config = useRuntimeConfig()
   const nuxtApp = useNuxtApp()
 
   return nuxtApp.$img as $Img || nuxtApp._img || (nuxtApp._img = createImage({
     ...imageOptions,
+    event: event || nuxtApp.ssrContext?.event,
```
pi0, line 14: This looks a little bit confusing. If we do expect `nuxtApp` context to be available (L10), and assuming nuxt app always has an event in its context, why it has to be passed to `useImage`?

src/runtime/composables.ts
```diff
@@ -1,15 +1,17 @@
...
 import type { $Img } from '../module'
 
 import { createImage } from './image'
 import { imageOptions } from '#build/image-options.mjs'
 import { useNuxtApp, useRuntimeConfig } from '#imports'
 
-export const useImage = (): $Img => {
+export const useImage = (event?: H3Event): $Img => {
   const config = useRuntimeConfig()
   const nuxtApp = useNuxtApp()
```
pi0, line 10: We are emulating `nuxtApp` for server no? Would be nice to typecase it to limited type (only `baseURL` is available rest are pertial) to avoid future assumptions

danielroe: yes, we already do this:

https://github.com/nuxt/image/blob/9a5e4965c826997658215b298f80575ab96255b2/src/types/image.ts#L40-L42

we don't use nuxtApp in the server (we only call `createImage` for nitro endpoints) - nuxtApp is only used in the vue version of `useImage`

danielroe (approved):

PR #1618, opened 2024-12-05
fix: set up non-default `ipx` providers if `options.ipx` is set

### LINKED ISSUE #1414: IPX provider not working properly when the default provider is not IPX

whatcould:
Fixes #1414

### ❓ Type of change

- [ x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Previously, specifying a custom provider would prevent setup of ipx, which happens in a `'nitro:init` hook. That is, there was no way use ipx if it wasn't your default provider.

Now, you just need to add `ipx` to the image options:

```
image: {
  ipx: {},
  provider: "yourprovider",
  providers: {
    yourprovider: { ... }
  }
}
```

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -134,13 +134,20 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
     })
 
     nuxt.hook('nitro:init', async (nitro) => {
-      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic' || 'ipx' in options || 'ipxStatic' in options) {
+      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic' || 'ipx' in options) {
         const resolvedProvider = nitro.options.static || options.provider === 'ipxStatic'
           ? 'ipxStatic'
           : nitro.options.node ? 'ipx' : 'none'
 
-        imageOptions.provider = options.provider ||= resolvedProvider
-        options[resolvedProvider] = options[resolvedProvider] || {}
+        if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic') {
+          imageOptions.provider = options.provider = resolvedProvider
+          options[resolvedProvider] = options[resolvedProvider] || {}
+        }
+
+        // handle the case of `ipx: {}` existing in options, but deploying a static site
+        if (resolvedProvider === 'ipxStatic' && options.provider !== resolvedProvider) {
+          options.ipxStatic ||= options.ipx
+        }
 
         const p = await resolveProvider(nuxt, resolvedProvider, {
           options: options[resolvedProvider],
--- a/src/module.ts
+++ b/src/module.ts
@@ -141,12 +141,15 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
 
         if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic') {
           imageOptions.provider = options.provider = resolvedProvider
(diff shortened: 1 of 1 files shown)


src/module.ts
```diff
@@ -134,12 +134,12 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
...
 
     nuxt.hook('nitro:init', async (nitro) => {
-      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic') {
+      if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic' || 'ipx' in options || 'ipxStatic' in options) {
         const resolvedProvider = nitro.options.static || options.provider === 'ipxStatic'
           ? 'ipxStatic'
           : nitro.options.node ? 'ipx' : 'none'
 
-        imageOptions.provider = options.provider = resolvedProvider
+        imageOptions.provider = options.provider || resolvedProvider
```
danielroe, line 142:
```suggestion
        imageOptions.provider = options.provider ||= resolvedProvider
```

danielroe:
Pushed a slight refactor - does that still work for you?

whatcould:
Seems fine? I couldn't quite figure out what ipxStatic did or make it work myself — though it does seem to require the ipx initialization, so if you wanted to use it alongside another default, your change would make it impossible.

danielroe (approved):

danielroe:
`ipxStatic` is basically `ipx` but only referencing files that were generated at build time, without any runtime component. And it should work fine - users can provide config in `ipx: {}` that gets passed to the `ipxStatic` provider.

src/module.ts
```diff
@@ -134,13 +134,20 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
...
 
-        imageOptions.provider = options.provider = resolvedProvider
-        options[resolvedProvider] = options[resolvedProvider] || {}
+        if (!options.provider || options.provider === 'ipx' || options.provider === 'ipxStatic') {
+          imageOptions.provider = options.provider = resolvedProvider
+          options[resolvedProvider] = options[resolvedProvider] || {}
+        }
+
+        // handle the case of `ipx: {}` existing in options, but deploying a static site
+        if (resolvedProvider === 'ipxStatic' && options.provider !== resolvedProvider) {
```
danielroe, line 148:
```suggestion
        if (resolvedProvider === 'ipxStatic') {
```

src/module.ts
```diff
@@ -134,13 +134,20 @@ ${providers.map(p => `  ['${p.name}']: { provider: ${p.importName}, defaults: ${
...
+          options.ipxStatic ||= options.ipx
```
danielroe, line 149:
```suggestion
        if (resolvedProvider === 'ipxStatic') {
          options.ipxStatic ||= options.ipx || {}
```

IO-Fire:
Fix #1676

PR #1626, opened 2024-12-10
feat(nuxt-img): add `custom` slot for full control of rendering

### LINKED ISSUE #1307: Placeholder slot for `<NuxtImg>`
I've been thinking of how to achieve more sophisticated placeholders using `<NuxtImg>` than what the `placeholder` prop allows for. Being able to define the placeholder in a slot would allow for greater flexibility:
```
<NuxtImg>
  <template #placeholder>
    <SkeletonLoader />
  </template>
</NuxtImg>
```
If this is a desirable addition, I am willing to implement it.

danekslama:
### 🔗 Linked issue

resolves #1307

### ❓ Type of change

### 📚 Description

This PR add a placeholder slot for NuxtImg component


docs/content/2.usage/1.nuxt-img.md
```diff
@@ -310,6 +312,12 @@ const nonce = useNonce()
 </script>
 ```
 
+## Slots
+
+### `placeholder`
+
+Use this slot to display a placeholder before the actual image loads
```
Baroshem, line 319: suggestion: maybe we could add an actual code example how to use this slot and how a placeholder could look like (it can be very simple one liner but still something that users can easily copy and paste into their project and adjust later on)

docs/content/2.usage/1.nuxt-img.md
```diff
@@ -100,6 +100,8 @@ To generate special versions of images for screens with increased pixel density.
 
 Display a placeholder image before the actual image is fully loaded.
 
+You may also use the [placeholder slot](/usage/nuxt-img#placeholder-1) to pass complex data, if needed.
```
Baroshem, line 103: question: is this link correct?

danekslama: It should be correct. I've tried it again to make sure it works and it does scroll to the correct position.

docs/content/2.usage/1.nuxt-img.md
danekslama: I added a snippet

```html
<NuxtImg
  src="/images/colors.jpg"
>
  <template #placeholder>
    I am visible before the image loads!
  </template>
</NuxtImg>
```

src/runtime/components/NuxtImg.vue
```diff
@@ -1,18 +1,24 @@
...
     v-bind="{
       ...isServer ? { onerror: 'this.setAttribute(\'data-error\', 1)' } : {},
       ...imgAttrs,
       ...attrs,
     }"
     :src="src"
   >
+
+  <slot
+    v-if="!isImageLoaded && $slots.placeholder"
```
danielroe, line 15:
```suggestion
    v-else
```

danekslama: @danielroe I don't think this would work, I used v-show on the img tag so that it always renders on server side.

danielroe: ahhh, quite right

Baroshem (approved):
Nice work! :)

src/runtime/components/NuxtImg.vue
danielroe: seems like this is a vue ssr bug, right? maybe instead use `isImageLoaded`? can we avoid checking `$slots.placeholder`?

danekslama: I am not sure what you mean by Vue SSR bug. I am using v-show because I need the element to be prepared on the server (it renders the  with the placeholder src or the desired image, or it renders the slot immediately on the server). If I use v-if, then the onload event won't be triggered because the element doesn't exist in the DOM. We can probably avoid checking $slots.placeholder.

danekslama: I simplified the condition

src/runtime/components/NuxtImg.vue
```diff
@@ -1,18 +1,24 @@
...
+    :class="props.placeholder && !isImageLoaded ? props.placeholderClass : undefined"
     v-bind="{
       ...isServer ? { onerror: 'this.setAttribute(\'data-error\', 1)' } : {},
       ...imgAttrs,
       ...attrs,
     }"
     :src="src"
   >
+
+  <slot
```
danielroe, line 14: my only remaining question is about possible breaking changes.

this makes `<NuxtImg>` into a non-single-root-element component, so (for example) it won't work as the child of a `<Transition>`. (That's why you had to update the tests to add `imageWrapper = wrapper.find('img')`. This would also have to be updated in any project using `NuxtImg`....)

It's also conceptually a change from the idea that `<NuxtImg>` would always just render an `` ...

I'd like to think about whether it's possible to implement this differently - maybe instead, like `<NuxtLink>`, support a `custom` prop that passes down the necessary pieces?

danekslama: I agree with it. I will look into it after Christmas.

danekslama: Should I make another issue or resolve it here?

danielroe: happy for you to resolve it here

danekslama:
I added the custom prop which acts similarly to the NuxtLink custom prop. I also updated the docs to show how to create custom placeholders with it.

danielroe (commented):
Great - this is very promising!

Would you be able to add some tests? 🙏

danekslama:
I'll be honest, I'm fairly new to unit testing, but I have some experience with Cypress. I added a test to check if the placeholder functionality and attribute fallthrough work as expected when the custom prop is set, and it seems they do.

danekslama:
@danielroe is this okay or should I add more tests?

src/runtime/components/NuxtImg.vue
```diff
@@ -1,14 +1,28 @@
...
+    v-else
+    v-bind="{
+      ...isServer ? { onerror: 'this.setAttribute(\'data-error\', 1)' } : {},
+      imgAttrs: {
+        ...imgAttrs,
+        ...attrs,
+      },
+      isLoaded: isImageLoaded,
```
danielroe, line 22: this looks perfect now, just one change - I think let's not expose `isImageLoaded` as this can be misleading - on server load the load may take place before hydration and this will never end up being set to `true`

danielroe: if it's significant, we could also open a new PR to implement the functionality - maybe through an approach similar to error handling?

danielroe (approved):

PR #1632, opened 2024-12-13
docs: make clearer that `xxl` is not a tailwind size

razbakov:
### 🔗 Linked issue

resolves https://github.com/nuxt/image/issues/1633

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

Tailwind has xl and 2xl
https://tailwindcss.com/docs/responsive-design

### DIFF
--- a/docs/content/1.get-started/2.configuration.md
+++ b/docs/content/1.get-started/2.configuration.md
@@ -71,6 +71,7 @@ export default defineNuxtConfig({
       'md': 768,
       'lg': 1024,
       'xl': 1280,
+      'xxl': 1536,
       '2xl': 1536
     },
   }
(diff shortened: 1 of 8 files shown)


Baroshem:
Hey there,

In the future, I think it would be better to create an issue first and then create a PR with a change 🙏

I am not sure if we should do that as this change is dictated by Tailwind while NuxtImg can work on their own without it. Could you elaborate why this should be added/removed from NuxtImg?

razbakov:
Hi @Baroshem `xxl` and `2xl` is the same thing. Also example contradicts what's written in the [docs](https://image.nuxt.com/get-started/configuration#screens):

> We share the same naming and sizes as [Tailwind CSS](https://tailwindcss.com/docs/responsive-design) and we added the xs property.

Either this message should be adjusted - I see no reason for that.
Or the extra property should be removed.

There are 2 ways to name sizes:
- xl
- xxl
- xxxl

Or: 
- xl
- 2xl
- 3xl

Counting letters is more difficult than reading the number.

Baroshem:
I see. Thanks for clarification! 

Then I think it makes sense but I wonder if it wouldn't be a breaking change? If someone was using current `xxl` size? @danielroe what do you think?

danielroe (approved):

razbakov:
@danielroe I would remove it from the config example in docs.

PR #1640, opened 2024-12-17
fix: emit error events from `<NuxtImg>` placeholder

### LINKED ISSUE #1639: When using Placeholder, the ERROR event will not be triggered
I need to deal with the ERROR event when the picture is loaded, but if I use the Placehoder attribute at the same time, the ERROR event will not be thrown out

nuxt-img.vue
```javascript
  mounted() {
    if (this.nPlaceholder) {
      const img = new Image();
      img.src = this.nMainSrc;
      // !!! Here, if the IMG loading error, there is no ERROR incident !!!
      img.onload = () => {
        this.$refs.img.src = this.nMainSrc;
        this.placeholderLoaded = true;
      };
    }
    if (process.server && process.static) {
      if (this.sizes) {
        this.nSizes;
      }
    }
  }
```

jianxing-xu:
resolves #1639 

### 📚 Description

I need to deal with the ERROR event when the picture is loaded, but if I use the Placehoder attribute at the same time, the ERROR event will not be thrown out

```javascript
onMounted(() => {
  if (placeholder.value) {
    const img = new Image()

    if (mainSrc.value) {
      img.src = mainSrc.value
    }

    if (props.sizes) {
      img.sizes = sizes.value.sizes || ''
      img.srcset = sizes.value.srcset
    }
    // !!! Here, if the IMG loading error, there is no ERROR incident !!!
    img.onload = (event) => {
      placeholderLoaded.value = true
      emit('load', event)
    }
    // add this to fix
    img.onerror = (event) => {
      emit('error', event)
    }

    markFeatureUsage('nuxt-image')

    return
  }
  // ...other
}
```

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -152,6 +152,10 @@ onMounted(() => {
       emit('load', event)
     }
 
+    img.onerror = (event) => {
+      emit('error', event)
+    }
+
     markFeatureUsage('nuxt-image')
 
     return
--- a/test/unit/image.test.ts
+++ b/test/unit/image.test.ts
@@ -160,15 +160,20 @@ describe('Renders simple image', () => {
 
 const getImageLoad = (cb = () => {}) => {
   let resolve = () => {}
+  let reject = () => {}
   let image = {} as HTMLImageElement
   const loadEvent = Symbol('loadEvent')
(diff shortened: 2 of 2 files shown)


Baroshem (approved):
Looks good to me! :)

danielroe (approved):

PR #1680, opened 2025-01-16
feat: add new provider filerobot

### LINKED ISSUE #1380: New provider for Filerobot from Scaleflex
At Malt we're using [Filerobot](https://docs.scaleflex.com/filerobot-documentation) for handling our images, which is the Digital Asset Management from Scaleflex (the same company that's developing and maintaining Cloudimage). We can basically apply the same common transformations between Filerobot and Cloudimage, such as width, height, fit, etc.

With `nuxt-image`, we can't really use the cloudimage provider because it requires a `token` in addition of the `baseURL`, which we don't have with Filerobot. We only need a `baseURL` (see for example: https://dam.malt.com/cms-front/maltLogo.png).

For the moment, we're using a custom provider but I was wondering if we could add it to the list of providers, directly within `nuxt-image`.

I'd gladly open a PR for this.

nicolaspayot:
### 🔗 Linked issue

Resolves #1380 

### ❓ Type of change

### 📚 Description

This PR adds a new provider (Filerobot from Scaleflex) to the list.


danielroe (approved):
I think we likely need an image for the provider, in `docs/public/providers/filerobot.svg`

PR #1686, opened 2025-01-26
fix: double loading image when placeholder is present

### LINKED ISSUE #1683: Double image loaded with placeholder
When adding placeholder on playground example

https://github.com/nuxt/image/blob/main/playground/pages/provider/%5Bprovider%5D.vue#L11-L15
as this
`
<nuxt-img
  :provider="provider.name"
  v-bind="sample"
  :preload="{ fetchPriority: 'auto' }"
  :placeholder="10"
/>
`
I'm getting double main image loaded ass seen in the screenshot

Image

george-moongroup:
### 🔗 Linked issue

resolves #1683

### ❓ fix

### 📚 Description

We resolved an issue where images could load more than once in the NuxtImg component. Previously, when a placeholder image was in use, it triggered a new image load even though the main image had already started loading. The update ensures that if a placeholder is already handled, we avoid loading the image again. This change prevents duplicate network requests and avoids unnecessary load events

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -150,7 +150,7 @@ const nuxtApp = useNuxtApp()
 const initialLoad = nuxtApp.isHydrating
 
 onMounted(() => {
-  if (placeholder.value || props.custom) {
+  if (props.custom) {
     const img = new Image()
 
     if (mainSrc.value) {
@@ -185,11 +185,14 @@ onMounted(() => {
       emit('error', new Event('error'))
     }
     else {
+      placeholderLoaded.value = true
       emit('load', new Event('load'))
     }
+    return
   }
 
   imgEl.value.onload = (event) => {
+    placeholderLoaded.value = true
     emit('load', event)
   }
 


src/runtime/components/NuxtImg.vue
```diff
@@ -150,7 +150,7 @@ const nuxtApp = useNuxtApp()
 const initialLoad = nuxtApp.isHydrating
 
 onMounted(() => {
-  if (placeholder.value || props.custom) {
+  if (props.custom) {
```
danielroe, line 153: there are things in this block we want to apply for placeholders, I think

george-moongroup: at this point, if SSR nor SSR the placeholder it is already loaded, if the 
`if (placeholder.value ...` will be kept a weird dance will happend the existing/loaded placeholder will get replaced by the mainSrc but the computed will repace it with the placeholder and then with the mainSrc once again ... so from here the double loading

pascalvaccaro: Following up on this, as we encountered the same issue using the `placeholder` prop. 

@george-moongroup Daniel is right, there are some usecases where the placeholder is needed in the `onMounted` callback, especially in non-SSR mode (the `placeholderLoaded` ref would never be updated, leaving the image `src` as the URL of the placeholder, not the main image)

@danielroe even with [a valid fix](https://github.com/pascalvaccaro/nuxt-image/pull/2/files#diff-64db07936246a3092bc10daa47f35ca097acb46b58a477441c37fd0f15af93d3) (I'm not asking to merge, the PR is for testing purposes only so I add to tweak the code to enable & run e2e testing), I'm still left wondering about the `preload` attribute in SSR-enabled environment when `placeholder` holds a "truthy" value: 
- should the `link` tag preload the placeholder URL? => it seems counter-intuitive as the goal of the placeholder is to "wait until" the main image is loaded. However the page will load faster this way as the browser won't have to wait for the main image.
- should the `link` tag preload the actual URL? => it seems a bit useless to have a placeholder if the main image is preloaded. However, the page will load slower because it'll have to wait for the main image to be loaded.

Another point could be made about the `loading` prop when set to "lazy": should the placeholder feature still run? The browser won't load the main image until it estimates that it will be needed imminently: I feel like whenever `loading="lazy"`, fetching both the placeholder AND the main image is a bit much, as the browser could only load the main image lazily.

Please advise 🙏  Thank you for your work! ❤️

vsergiu93: FYI: when placeholder holds a truthy value, loading="lazy" does not behave as expected.
All images are loaded immediately, regardless of how far they are from the viewport.

I'm seeing this in a production application, and I just cloned the repo and confirmed the behavior in the playground.

I agree with @pascalvaccaro that the interaction between `placeholder`  and  `loading="lazy"`  needs clarification. 

@danielroe Let me know if you want me to create a PR where I skip the logic of placeholder when `loading="lazy"` is set.
Thanks

Issue #442, opened 2021-10-20
Generating WebP images with nuxt-picture not working for static generated site (ssr:true,target:static)


jakubferenc:
Nuxt version: 2.15.8
@nuxt/image version: 0.5.0
target: 'static', 
ssr: true,
buildModules: [
  ['@nuxt/image'],
],
image: {
  provider: 'static',
},

Using nuxt-picture with WebP type works in the dev mode. It generates images via _ipx. Here's a sample code from the dev mode that works:

This code (in dev mode)
```
  nuxt-picture.image.image-desktop(
    format="webp"
    quality="80"
    src="/images/xyz-header-large.jpg"
    sizes="mobile:100vw tablet:100vw desktop:100vw widescreen:100vw fullhd:100vw"
  )
```

will generate this code, which uses WebP images (in dev mode)

`<source type="image/webp" srcset="/_ipx/images/xyz-header-large.jpg?w=768&amp;f=webp&amp;q=80 768w, /_ipx/images/xyz-header-large.jpg?w=769&amp;f=webp&amp;q=80 769w, /_ipx/images/xyz-header-large.jpg?w=1024&amp;f=webp&amp;q=80 1024w, /_ipx/images/xyz-header-large.jpg?w=1216&amp;f=webp&amp;q=80 1216w, /_ipx/images/xyz-header-large.jpg?w=1408&amp;f=webp&amp;q=80 1408w" sizes="(max-width: 768px) 100vw, (max-width: 769px) 100vw, (max-width: 1024px) 100vw, (max-width: 1216px) 100vw, 100vw">`

However, when trying to generate a static site using yarn generate, the console output indicates that WebP images were created and indeed I can find them in **_nuxt/image/** folder. But the generated static code does not link to WebP images, but uses the original "jpg" or "png" images instead. 

Any help on this would be greatly appreciated.

shanks25:
same issue here

GuillaumeDgr:
Hi,
Any update on this pls ?
Many thks

bropa:
Same problem here. 
I looked at the whole thing in Chrome dev-tools. Both variants (jpg/webp) are delivered from different folders:
`http://localhost:3000/_ipx/q_75&w_1920&f_jpeg/referenzen/placeholder-lrg.jpg`
`http://localhost:3000/_ipx/q_75&w_1920&f_webp/referenzen/placeholder-lrg.jpg `
[...]

PR #1793, opened 2025-04-04
feat(ipx): use alternate URL style with format as file extension

### LINKED ISSUE #933: doesn't generate other formats on "nuxt generate"
Working with
```json
"@nuxt/image-edge": "^1.0.0-rc.1-28164352.ad7c010",
"nuxt": "^3.6.5",
```

```js
image: {
    format: ['webp'],
    provider: 'ipx',
    ipx: {
        modifiers: {
            quality: '80',
            format: ['webp']
        }
    }
}
```

tried on macbook air m2 and netlify builds. The images are transformed but no other formats are generated.

sgarner:
### 🔗 Linked issue

Resolves #933, #584, #442

### ❓ Type of change

- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Implements (and depends on) the alternate URL style for IPX proposed in https://github.com/unjs/ipx/pull/259 

This improves prerendering of images so that the target format is respected in the generated filename, fixing issues with static hosting. For full rationale please see the description in https://github.com/unjs/ipx/pull/259 

The original URL style will still be used when the requested format is `auto` (which is not applicable for prerendering anyway), or when no modifiers are given. Otherwise the new URL style will be used.

I'm not sure whether this needs to be classed as a breaking change. It will cause different URLs to be produced with `ipx` and `ipxStatic` providers, but that shouldn't break anything _unless_ users are somehow relying on the old URL style.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "globby": "^14.1.0",
     "happy-dom": "^17.4.4",
     "installed-check": "^9.3.0",
-    "ipx": "^2.1.0",
+    "ipx": "https://github.com/sgarner/ipx/releases/download/v3.0.4-alpha.feat-format-extension.1/ipx-3.0.4-alpha.feat-format-extension.1.tgz",
     "jiti": "2.4.2",
     "knip": "^5.46.5",
     "nitropack": "^2.11.8",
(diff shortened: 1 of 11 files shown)


hechang27-sprt:
@danielroe please review this as this will save so much confusion for new users who don't know that neither simple http servers like the npx serve nor the dev tool will detect actual image type with magic bits etc.. They only look content-type and file extensions.

PR #1802, opened 2025-04-05
feat!: typed providers + modifiers

danielroe:
### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This adds strong typing for image providers + modifiers throughout, including in the image module options. (For example, a provider that requires you set a baseURL will enforce that in the nuxt config.

This is a breaking change because image providers are now no longer using named exports but a single default export wrapped in a `defineProvider` function (which can also have an optional setup function).

I've also done quite a lot of cleaning up of types elsewhere, and dropped image props that we weren't using - they should be fine to be attributes inherited by the `` element.

The next steps will need to be actually getting and setting the types for provider modifiers. (As an interim fix we can support key signatures `[key: string]: string | number` that would allow the same permissive approach that's working at the moment.)

Because types live in the individual provider files, it should be easy for users to add their own types for custom providers - and also to keep the types in sync for future providers + updates to existing ones.


atinux (approved):
Really good changes IMO, I like the `defineProvider` util!

I let you fix conflicts with pnpm lock before merging of course 😄

PR #1808, opened 2025-04-08
feat(ipx): log the architecture of the build

danielroe:
### ❓ Type of change

### 📚 Description

With upcoming release of nuxt image including ipx v3, it's important to let users know the architecture they have built for.

This adds a log after the build is complete.

CleanShot 2025-04-08 at 09 38 23@2x

questions:
1. is it that important? we could move it earlier in the build pipeline rather than making sure it's the last thing users see
2. do we need to do more than add this log? the main thing is that it's there for is to help users debugging issues they might face

### DIFF
--- a/src/ipx.ts
+++ b/src/ipx.ts
@@ -1,6 +1,8 @@
+import { arch, platform } from 'node:os'
+import { readdir } from 'node:fs/promises'
 import { existsSync } from 'node:fs'
-import { relative, resolve } from 'pathe'
-import { useNuxt, createResolver, useNitro } from '@nuxt/kit'
+import { join, relative, resolve } from 'pathe'
+import { useNuxt, createResolver, useNitro, useLogger } from '@nuxt/kit'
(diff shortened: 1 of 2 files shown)


atinux:
I like the idea of explaining regarding the binaries which are shipped in the output.

Do we know what is the error sent back by the IPX route when the binary is not build for the same runtime?

danielroe:
@atinux The response when loading an image is a 500 error and:

```json
{
  "error": {
    "message": "[500] [IPX_ERROR] Could not load the \"sharp\" module using the darwin-arm64 runtime\nPossible solutions:\n- Ensure optional dependencies can be installed:\n    npm install --include=optional sharp\n- Ensure your package manager supports multi-platform installation:\n    See https://sharp.pixelplumbing.com/install#cross-platform\n- Add platform-specific dependencies:\n    npm install --os=darwin --cpu=arm64 sharp\n- Consult the installation documentation:\n    See https://sharp.pixelplumbing.com/install"
  }
}
```

Issue #438, opened 2021-10-13
Expose a reference of the underlying img HTMLElement


Rigo-m:
### Proposal
`<nuxt-img />` and `<nuxt-picture />` should expose in some way the underlying `<img />` HTMLElement.

### Why would I need this?
As pointed out here: [Imgix blurhash issue](https://github.com/nuxt/image/pull/364#issuecomment-942037622), I'm displaying a blurhashed image on top of `<nuxt-img />`, and I need to un-show it either if the image has been loaded (i.e. it's already present because of cache) or after the `onload` event.
I've attached a `ref` to `<nuxt-img />` and then I've searched for the `<img />` element underneath to check `image.complete` attribute, but it feels a little bit clunky.
How can we handle this better?

everyx:
I also have the same problem, it seems to be that the `nuxt-img` component in SSR mode will not emit `onload` when the page first load.

nathanchase:
I have the same issue. I'd like to use the `placeholder` attribute of `NuxtImg`, but then have the actual loaded image fade in when it has loaded, rather than "pop" in over the placeholder.

@Rigo-m How did you reference the img inside NuxtImg?

I tried this, but it's not quite right:
```
const nuxtImageRef = ref<HTMLElement | null>(null);

const doSomethingOnLoad = () => {
  const imgElement = nuxtImageRef.value?.querySelector('img');

  if (imgElement) {
    imgElement.classList.add('loaded');
  }
};
```

Rigo-m:
Hi Nathan, what version are you using of nuxt and nuxt image?

nathanchase:
> Hi Nathan, what version are you using of nuxt and nuxt image?

nuxt: 3.8.0
@nuxt/image: 1.0.0

nathanchase:
Seems like Vue 3's `defineExpose` might be a possibility? https://vuejs.org/api/sfc-script-setup.html#defineexpose

To explicitly expose properties in a <script setup> component, use the defineExpose compiler macro:

```
<script setup>
import { ref } from 'vue'

const a = 1
const b = ref(2)

defineExpose({
  a,
  b
})
</script>
```

> When a parent gets an instance of this component via template refs, the retrieved instance will be of the shape { a: number, b: number } (refs are automatically unwrapped just like on normal instances).

Perhaps `<NuxtImg>` can expose a ref to the `img` this way?

LeonardoRick:
https://github.com/nuxt/image/pull/1834
 
[...]

PR #1834, opened 2025-05-10
expose image element on nuxtImg through defineExpose

LeonardoRick:
### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Expose the image element as its currently impossible to access it without a `querySelector`


test/nuxt/image.test.ts
```diff
@@ -153,6 +153,17 @@ describe('Renders simple image', () => {
     const domNonce = img.element.getAttribute('nonce')
     expect(domNonce).toBe('stub-nonce')
   })
+
+  it.only('can access imgEl underlying img HTMLImageElement', () => {
```
danielroe, line 157:
```suggestion
  it('can access imgEl underlying img HTMLImageElement', () => {
```

danielroe (commented):
what do you want to do with the image element?

LeonardoRick:
@danielroe In my specific case I want to rotate the image in realtime. The `rotate` modifier was not working and I see [some people](https://github.com/nuxt/image/issues/438) also wanting to access it. 

My solution for now is using `querySelector('img')` but would be nice to access the element directly.


```js
  () => {
    if (!self.value) return;
    const img = self.value.querySelector('img');
    if (!img) return;
    img.style.setProperty('--background-rotation', `${rotate}deg`);
  }
```

LeonardoRick:
Hey! Any reason not to approve this one?

PR #1836, opened 2025-05-11
feat: automatically add crossorigin attribute to preload link

### LINKED ISSUE #1835: Preload add crossorigin attribute
I have an image that I want to preload, that comes from an external source.

That's why I have added the `crossorigin="anonymous"` and `referrerpolicy="no-referrer"` attributes.

But now the generated HTML is only the preload without those extra attributes:

```html
<link rel="preload" as="image" href="XXX">
```

Because of that, Chrome throws following warning in the console:

```
A preload for 'XXX' is found, but is not used because the request credentials mode does not match.
Consider taking a look at crossorigin attribute.
```

So the generated HTML should essentially look like this:

```html
<link rel="preload" as="image" href="XXX" crossorigin="anonymous" referrerpolicy="no-referrer">
```

sotasan:
### 🔗 Linked issue

Resolves #1835 

### ❓ Type of change

### 📚 Description

It fixes the following Chrome warning when setting a crossorigin attribute for external images:

```
A preload for 'XXX' is found, but is not used because the request credentials mode does not match.
Consider taking a look at crossorigin attribute.
```

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -121,6 +121,7 @@ if (import.meta.server && props.preload) {
       ...(typeof props.preload !== 'boolean' && props.preload.fetchPriority
         ? { fetchpriority: props.preload.fetchPriority }
         : {}),
+      crossorigin: normalizedAttrs.value.crossorigin,
     }],
   })
 }


danielroe (approved):
thank you! ❤️ 

my apologies for the long delay on this.

PR #1884, opened 2025-07-01
fix(NuxtImg): add image onerror handler in internal script to avoid s…

### LINKED ISSUE #1011: Support for nonce or hash value [enhancment]

bobeatspie:
…cript-src-attr CSP violation

### 🔗 Linked issue

Resolves #1011

### ❓ Type of change

- [x ] 👌 Enhancement (improving an existing functionality)

### 📚 Description

The CSP header "script-src-attr" prevents inline javascript from executing when the header's value is set to 'none'. By adding the error event handler in the onMounted hook, it becomes easier for Nuxt Security to add a hash to the script and allow the error handler to execute.

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -63,7 +63,6 @@ const imgAttrs = computed(() => ({
   ...(!props.placeholder || placeholderLoaded.value)
     ? { sizes: sizes.value.sizes, srcset: sizes.value.srcset }
     : {},
-  ...import.meta.server ? { onerror: 'this.setAttribute(\'data-error\', 1)' } : {},
   ...attrs,
 }))
 
@@ -133,6 +132,10 @@ if (import.meta.server && import.meta.prerender) {
 const initialLoad = useNuxtApp().isHydrating
 const imgEl = useTemplateRef('imgEl')
 onMounted(() => {
+  if (imgEl.value) {
+    imgEl.value.addEventListener('error', handleImgError)
+  }
+
   if (placeholder.value || props.custom) {
     const img = new Image()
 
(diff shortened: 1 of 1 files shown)


danielroe:
the purpose of the inline js is to capture errors that occur before vue has hydrated the page

we might also be able to add a script in `<head>` that adds an event listener for image load events....

I've marked as draft for my own triage - feel free to unmark whenever it's ready for review.

maximtrp:
@danielroe Did you mean something like this?
```js
useHead({
  script: [{
    key: 'nuxt-image-error-listener',
    tagPosition: 'head',
    nonce: props.nonce,
    innerHTML:
      "document.addEventListener('error',function(e){var t=e.target;" + 
      "if(t&&t.tagName==='IMG'&&(t.hasAttribute('data-nuxt-img')||t.hasAttribute('data-nuxt-pic')))" +
      "t.setAttribute('data-error','1')},true)",
  }],
})
```

PR #1984, opened 2025-10-14
fix: double loads of the original image when placeholder true

### LINKED ISSUE #1683: Double image loaded with placeholder

amrelbialy:
🔗 Linked issue

resolves #1683

## ❓ Fix

- [ ]  📖 Documentation (updates to the documentation or readme)
- [x]  🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ]  👌 Enhancement (improving an existing functionality)
- [ ]  ✨ New feature (a non-breaking change that adds functionality)
- [ ]  🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ]  ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

📚 Description
Resolved an issue where images could load more than once in the NuxtImg component. Previously, when a placeholder image was in use, it triggered a new image load even though the main image had already started loading. The update ensures that if a placeholder is already handled, we avoid loading the image again. This change prevents duplicate network requests and avoids unnecessary load events. By removing the preload using new image() instance and adding onLoad to imgAttrs, and using vue reactive to handle the update when placeholderLoaded is true

### DIFF
--- a/playground/app/pages/placeholder-regression.vue
+++ b/playground/app/pages/placeholder-regression.vue
@@ -1,0 +1,10 @@
+<template>
+  <div>
+    <nuxt-img
+      src="/images/colors.jpg"
+      placeholder
+      width="500"
+      height="500"
+    />
+  </div>
+</template>
--- a/test/e2e/ssr.test.ts
+++ b/test/e2e/ssr.test.ts
@@ -84,4 +84,28 @@ describe('browser (ssr: true)', () => {
       }
     `)
   })
+
+  it('should not load the main image twice when placeholder is enabled', async () => {
+    const page = await createPage()
+
+    const requests: string[] = []
+    await page.route('**', (route) => {
+      requests.push(route.request().url())
+      return route.continue()
+    })
+
+    await page.goto(url('/placeholder-regression'), { waitUntil: 'networkidle' })
+
+    await page.waitForSelector('img')
+
+    const mainImageRequests = requests.filter(r =>
+      r.includes('/_ipx/')
+      && r.includes('images/colors.jpg')
+      && r.includes('s_500x500'),
+    )
+
+    expect(mainImageRequests.length).toBe(1)
+
+    await page.close()
+  })
 })


Aurelien-GH:
Hi @danielroe 
Can we have a look at this fix please, it's really needed on our side.
Thanks in advance for the feedback!

src/runtime/components/NuxtImg.vue
```diff
@@ -130,32 +160,6 @@ if (import.meta.server && import.meta.prerender) {
...
-    img.onload = (event) => {
-      placeholderLoaded.value = true
-      emit('load', event)
-    }
-
-    img.onerror = (event) => {
-      emit('error', event)
-    }
-
-    markFeatureUsage('nuxt-image')
```
danielroe, line 154: we want to keep this

amrelbialy: I think my solution replace this old part. it works the same but avoids having multiple calls, what do you think?

danielroe: I'm talking about `markFeatureUsage`

amrelbialy: ah ok i will add it again

amrelbialy: Added

danielroe:
would you be able to add a regression test case in `ssr.test.ts` to reproduce the bug and make sure it doesn't happen again? 🙏

amrelbialy:
@danielroe added

amrelbialy:
Hey @danielroe, any updates?

Aurelien-GH:
@danielroe let us know if anything else needed.

Aurelien-GH:
@danielroe feel free to tell us and @amrelbialy if anything missing, would love to have this behind us.

PR #2005, opened 2025-11-03
fix(ipx): encode ampersands in image uris

danielroe:
### 🔗 Linked issue

closes https://github.com/nuxt/image/pull/1268

### ❓ Type of change

### 📚 Description

this encodes ampersands in image uris for ipx

remake of https://github.com/nuxt/image/pull/1268 by @riddla as I don't have permission to push to that branch

### DIFF
--- a/src/runtime/providers/ipx.ts
+++ b/src/runtime/providers/ipx.ts
@@ -57,6 +57,7 @@ export const operationsGenerator = createOperationsGenerator({
     background: 'b',
     position: 'pos',
   },
+  joinWith: encodeURIComponent('&'),
   formatter: (key, val: string | number | boolean) => encodeParam(key) + '_' + encodeParam(val.toString()),
 })
 
@@ -70,7 +71,7 @@ export default defineProvider<Partial<IPXOptions>>({
       delete modifiers.height
     }
 
-    const params = encodeURIComponent(operationsGenerator(modifiers) || '_')
+    const params = operationsGenerator(modifiers) || '_'
 
     if (!baseURL) {
       baseURL = joinURL(ctx.options.nuxt.baseURL, '/_ipx')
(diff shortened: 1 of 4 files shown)


riddla:
@danielroe, thank you for your work. I integrated your refactoring (locally) into my branch. 

I think we might be encountering a bug in ipx – I pulled the latest ipx version locally and the 500 error seems to have to do with the order of the modifiers in the URL.

http://localhost:3001/embed%26f_webp%26s_400x400/picsum/id/237/600 does work, http://localhost:3001/q_20%26embed%26f_webp%26s_400x400/picsum/id/237/600 respectively using the quality modifier as the first modifier in the URL does trigger the 500 error (that you also encountered in the playground).

Screenshot 2025-11-04 at 09 51 49

(The error is only triggered, when I use the encoded ampersands, http://localhost:3001/q_20&embed&f_webp&s_400x400/picsum/id/237/600 is fine.)

Where should we go from here? File a bug over at https://github.com/unjs/ipx?

danielroe:
interesting. yes, in that case let's fix the issue in ipx

(or if it is as simple as sorting modifiers alphabetically we can do that in nuxt/image as a hotfix)

danielroe:
@riddla I investigated and the issue is that the decoding happens _after_ splitting by modifier, which must be either `&` or `,`. (and `,` can cause [issues within srcset](https://github.com/unjs/ipx/issues/57).

we could either open a PR to ipx to decode the entire string _first_, or possibly we might choose a different separator? @pi0 may have some thoughts here.

pi0:
if sorting can be a hotfix here would be great. in next major of IPX later we can change sep and make it configurable.

riddla:
Just for reference: I re-tested the Slack behavior regarding ampersands this morning with https://www.buescher.de/_ipx/fit_inside&f_jpg&s_1200x630/img/preview/home.jpg respectively https://www.buescher.de/_ipx/fit_inside%26f_jpg%26s_1200x630/img/preview/home.jpg.

There is still no preview for URLs using ampersands:
Screenshot 2025-11-04 at 11 04 01

PR #2007, opened 2025-11-04
docs: add migration guide

danielroe:
### ❓ Type of change

### DIFF
--- a/docs/content/1.get-started/5.migration.md
+++ b/docs/content/1.get-started/5.migration.md
@@ -78,7 +78,7 @@ export default defineNuxtConfig({
   image: {
     screens: {
       xs: 320,
-      xxl: 2560,
+      xxl: 1536
     }
   }
 })


docs/content/1.get-started/5.migration.md
```diff
@@ -0,0 +1,158 @@
...
+The `xs` and `xxl` breakpoints have been removed to align with Tailwind CSS defaults.
+
+Search for usage in your project:
+
+```bash
+grep -r "sizes.*xs:" --include="*.vue"
+grep -r "sizes.*xxl:" --include="*.vue"
+```
+
+**If you are using these screen sizes**, you can either replace `xs` with `sm` and `xxl` with `2xl`:
```
DamianGlowala, line 64: This holds for xxl, but not for xs.

See https://github.com/nuxt/image/pull/1931/files#diff-febc21f95a6b1803f5a608291db3659113139a4f37e9b58afa91cd9d9bfa600fL69-L75

docs/content/1.get-started/5.migration.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+... or you can add them back in your config:
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  image: {
+    screens: {
+      xs: 320,
+      xxl: 2560,
```
DamianGlowala:
```suggestion
      xxl: 1536
```

docs/content/1.get-started/5.migration.md
danielroe: yes, it's not the same value, but likely they should use sm: where they were previously using xs:

PR #2009, opened 2025-11-04
chore: drop node v18 support

iiio2:
Actually I think nuxt has dropped support for node v18. But here in the package.json file, the engines node is still v18. That is why I've updated it to node v20.


danielroe (commented):
this field is informative as far as what _nuxt/image_ requires, and all of its dependencies are fine with node 18.

we even have a test for it:

https://github.com/nuxt/image/blob/3e1f1f67f3ddee135ecdfeada224ccf6a64ea379/.github/workflows/ci.yml#L35-L36

if a user installs nuxt/image with a version of nuxt that requires a higher node version, they should get a notification from nuxt, not nuxt/image

iiio2:
@danielroe , ok thanks. then i close the pr.

PR #2016, opened 2025-11-06
fix: fix types for custom providers shadowing builtins

### LINKED ISSUE #2015: Bug: Custom provider types are not generated correctly if it has the same name as a built in provider
This was introduced with v2.0.0.

We have a custom shopify provider in our project.

The type generation returns the wrong type.

Image

Code line: https://github.com/nuxt/image/blob/main/src/module.ts#L125

dreitzner:
fix correct type generation for custom providers with same name as built in providers

### 🔗 Linked issue

resolves #2015 

### ❓ Type of change

### 📚 Description

This fixes the ttype generation for cases where custom providers have the same name as built in providers.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -3,7 +3,7 @@ import process from 'node:process'
 import { hasProtocol, parseURL, withLeadingSlash } from 'ufo'
 import { defineNuxtModule, addTemplate, addImports, addServerImports, createResolver, addComponent, addPlugin, addServerTemplate, addTypeTemplate } from '@nuxt/kit'
 import { join, relative, resolve } from 'pathe'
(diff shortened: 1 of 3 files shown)


danielroe (commented):
I wonder if we can resolve this with type safety by telling custom provider authors how to augment these interfaces

dreitzner:
> I wonder if we can resolve this with type safety by telling custom provider authors how to augment these interfaces

Not sure how I would implement that 😁

danielroe:
pushed a change - does this work in your use case?

dreitzner:
> pushed a change - does this work in your use case?

@danielroe yes that works and those changes look a little cleaner than mine 👍 

Thank you

PR #2026, opened 2025-11-12
chore: remove deprecated github silent config

iiio2:
github silent is now deprecated. I think we may remove it. 
https://vercel.com/docs/project-configuration/git-configuration#github.silent


danielroe:
thank you

iiio2:
Thanks @danielroe .

PR #2054, opened 2025-12-03
fix(bunny): merge query params with existing image src

### LINKED ISSUE #2055: bunny provider: can't handle src urls with query params
Bunny image provider can't handle URLs with query params correctly:

src: https://example.com?foobar

Result of bunny image provider: https://example.com?foobar?quality=80

Expected: https://example.com?foobar&quality=80

infabo:
In case the image src already contains a query parameter, it is necessary to properly add bunny query params. This is accomplished using withQuery helper from ufo.

### 🔗 Linked issue

resolves #2055

### ❓ Type of change

### 📚 Description

In case the image src already contains a query parameter, it is necessary to properly add bunny query params. This is accomplished using withQuery helper from ufo.

e.g. url is http://example.com/image.jpg?timestamp

### DIFF
--- a/src/runtime/providers/bunny.ts
+++ b/src/runtime/providers/bunny.ts
@@ -1,4 +1,4 @@
-import { joinURL } from 'ufo'
+import { withBase, withQuery, getQuery } from 'ufo'
 import { createOperationsGenerator } from '../utils/index'
 import { defineProvider } from '../utils/provider'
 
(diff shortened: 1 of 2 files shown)


src/runtime/providers/bunny.ts
```diff
@@ -1,9 +1,8 @@
-import { joinURL } from 'ufo'
-import { createOperationsGenerator } from '../utils/index'
+import { withBase, withQuery } from 'ufo'
+import { createMapper } from '../utils/index'
 import { defineProvider } from '../utils/provider'
 
-const operationsGenerator = createOperationsGenerator({
-  keyMap: {
+const operationsGenerator = createMapper({
```
danielroe: is there a reason you changed this? I think it's causing the type checking errors below

infabo: Because `withQuery` needs an mapped object instead of a concatenated query string.

PR #2084, opened 2026-01-12
feat: extend prop types from ImgHTMLAttributes

### LINKED ISSUE #2083: Type errors with img html attributes when `checkUnknownProps` is enabled
Hello,

If I turn on the [`checkUnknownProps` option in `vueCompilerOptions`](https://github.com/vuejs/language-tools/wiki/Vue-Compiler-Options#checkunknownprops-v222), the `NuxtImg` component seems to error out with multiple img html attributes such as `alt` and `loading`.

I think it would be nice to either add some more html attributes to the [explicitly defined props here](https://github.com/nuxt/image/blob/7eacf29ea25149b715622fdc3962e6e7746304a7/src/runtime/utils/props.ts#L27), or make the props extend from the `ImgHTMLAttributes` type. The same problem happens with `NuxtPicture` but that seems to be a little more complex (because certain specific properties seem to be manually passed down to the img elements).

gbyesiltas:
### 🔗 Linked issue

resolves #2083 

### ❓ Type of change

### 📚 Description

Makes the `NuxtImg` component props extend the `ImgHTMLAttributes` type so that the module can be used with a strict-prop typescript configuration. See [the issue](https://github.com/nuxt/image/issues/2083) for more details

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -29,7 +29,9 @@ import type { ProviderDefaults, ConfiguredImageProviders } from '@nuxt/image'
 
 import { useHead, useNuxtApp, useRequestEvent } from '#imports'
 
-export interface ImageProps<Provider extends keyof ConfiguredImageProviders> extends BaseImageProps<Provider> {
+export interface ImageProps<Provider extends keyof ConfiguredImageProviders>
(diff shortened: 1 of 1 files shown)


gbyesiltas:
looks like there are some issues with the build, closing the pr

Issue #2020, opened 2025-11-07
The density parameter in the configuration is not redefined, but adds new values ​​to the default ones


Semdevmaster:
NuxtImage [v2.0.0](https://github.com/nuxt/image/releases/tag/v2.0.0)
```vue
<NuxtImg src="/test.jpg" width="500"/>
```
```js
export default defineNuxtConfig({
  image: {
    densities: [4]
  }
})

```
This config gives this result
```html
<img width="500" 
data-nuxt-img
 srcset="
/_ipx/f_avif&amp;w_500&amp;q_80/test.jpg 1x,
/_ipx/f_avif&amp;w_1000&amp;q_80/test.jpg 2x,
/_ipx/f_avif&amp;w_2000&amp;q_80/test.jpg 4x"
 onerror="this.setAttribute('data-error', 1)"
 src="/_ipx/f_avif&amp;w_500&amp;q_80/test.jpg">
```
If I only want 1x and use this configuration
```js
export default defineNuxtConfig({
  image: {
    densities: [1]
  }
})
```
I still get the same result (1x adn 2x) but expected only (1x)
```html
<img width="500" 
data-nuxt-img
 srcset="
/_ipx/f_avif&amp;w_500&amp;q_80/test.jpg 1x,
/_ipx/f_avif&amp;w_1000&amp;q_80/test.jpg 2x"
 onerror="this.setAttribute('data-error', 1)"
 src="/_ipx/f_avif&amp;w_500&amp;q_80/test.jpg">
```
The only thing that helps is either using a preset or the density attribute in NuxtImg

It's a similar story with the format parameter
```js
export default defineNuxtConfig({
  image: {
    format: ['avif'],
  }
})
```
The image is still rendered as a JPEG, not an AVIF.
This can only be resolved by specifying a modifier for a specific provider or by using the format attribute in NuxtImg

julbd:
Related to https://github.com/nuxt/image/issues/1874

Issue #1874, opened 2025-06-21
Array/object default options (format, screen, domains, densities…) cannot be overrided


julbd:
Due to the usage of `DefineNuxtModule({ defaults: ... })`, the user-provided configuration in `nuxt.config.js` is merged with module defaults, including all array and object options.

For exampke, this behavior prevent the user to override default `densities: [ 1, 2 ]` with `densities: [ 1 ]`. This is a major issue to me.

wokalek:
Has anyone solved this with a pnpm patch?

PR #2138, opened 2026-02-25
fix!: allow overriding default `format`/`screens`/`densities`

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/image/issues/1874
resolves https://github.com/nuxt/image/issues/2020

### 📚 Description

This moves non-empty arrays and objects from `defaults` of `defineNuxtModule` to setup block, so that they are not affected by default merging behaviour.

### DIFF
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -16,8 +16,6 @@ export default defineNuxtConfig({
       'https://images.unsplash.com',
       'https://upload.wikimedia.org',
     ],
-    format: ['webp', 'jpg', 'jpeg'],
-    densities: [1, 2, 3, 4],
     screens: {
       'sm': 640,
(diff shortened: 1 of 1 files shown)


Chrissi2812:
Finally a fix for this 👍 This has been a blocker to upgrade to v2, for me.

But the `options.screens` would be nice if they still be merged by defu, and filtered via `.filter(Boolean)` so that we could write something like this:

```ts
  image: {
    screens: {
      'sm': undefined,
      '3xl': 2000,
    }
  }
})
```
But that's just my opinion ^^

danielroe:
this seems like a breaking change - if users were previously only _adding_ a value this will be unexpected.

we likely need to defer this to the next major of nuxt/image

PR #2140, opened 2026-02-25
fix(fastly): handle absolute URLs and merge preexisting query params

FADL285:
## Summary

- Replace `joinURL(baseURL, src + '?' + operations)` with `withBase(withQuery(src, parseQuery(operations)), baseURL)` in the Fastly provider
- Fixes absolute URLs being mangled when passed as `src` (e.g. `https://media.store.com/image.jpg`)
- Fixes existing query params being broken by double `?` (e.g. `src?v=123?width=300`)

Follows the same pattern used in #2135 (Shopify fix).

## Context

Magento/ecommerce backends return images from multiple CDN domains behind Fastly. The previous implementation broke when:
1. `joinURL('https://fastly.example.com', 'https://media.store.com/image.jpg')` → mangled URL
2. `src + '?' + operations` when `src` already contains `?v=123` → double `?`

> **Note:** ~15 other providers have the same bugs (imgix, gumlet, sirv, directus, etc.). Will open a separate tracking issue for those.

## Test plan

- [x] Updated Fastly test to use absolute URL with existing query params (mirrors Shopify test pattern)
- [x] Updated all 6 Fastly snapshot expectations in `test/providers.ts`
- [x] Updated bundle size snapshot (smaller due to `withBase`/`withQuery` vs `joinURL`)
- [x] `pnpm vitest run test/nuxt/providers.test.ts` — 43 tests pass
- [x] `pnpm lint` — clean


kheiner:
The pattern implemented into the providers.ts file feels like it will be problematic down the line.
https://github.com/nuxt/image/blob/860cd2d873a40ef2212dbb6e38de557436e58ceb/test/providers.ts#L11-L12

I'm working on improving the tests for provider-coverage #2192 and would like to get an understanding as to why `FastlyAbsoluteUrl` was necessary. (@FADL285)

I understand that you were working to mimic Shopify, however the Shopify implementation does not create a new key in /test/provider.ts:
https://github.com/nuxt/image/blob/860cd2d873a40ef2212dbb6e38de557436e58ceb/test/nuxt/providers.test.ts#L458-L470

Issue #2107, opened 2026-02-06
Unknown provider: supabase


marco-land:
When configuring the Supabase provider as described in the [documentation](https://image.nuxt.com/providers/supabase#configuration), TypeScript throws an error on the provider.

## Config
```ts
export default defineNuxtConfig({
  image: {
    supabase: {
      baseURL: 'https://<project-id>.supabase.co/storage/v1/render/image/public/<bucket-name>'
    }
  }
})
```

## Error
```bash
Uncaught (in promise) Error: Unknown provider: supabase
    at getProvider (image.js?v=517378ce:83:11)
    at resolveImage (image.js?v=517378ce:45:31)
    at getImage (image.js?v=517378ce:11:19)
    at $img (image.js?v=517378ce:17:48)
    at ComputedRefImpl.fn (NuxtImg.vue:73:41)
    at refreshComputed (reactivity.esm-bundler.js?v=517378ce:361:29)
    at get value (reactivity.esm-bundler.js?v=517378ce:1636:5)
    at NuxtImg.vue:101:17
    at runtime-core.esm-bundler.js?v=517378ce:3079:40
    at callWithErrorHandling (runtime-core.esm-bundler.js?v=517378ce:197:19)
```

danielroe:
shipped in #2027 - this was an issue where docs were visible before the release

Issue #2036, opened 2025-11-19
TypeScript error on useImage provider option with v2.0.0


fredericDelaporte:
The TypeScript definition of the function yielded by [`useImage`](https://image.nuxt.com/usage/use-image) has changed and now causes typing errors when willing to use its `provider` option.

That option was typed as `string`, but is now typed as only `'ipx'`.

The `ImageOptions` seems to have been changed for listing the configured providers there. But for some reason, that does not work in my project.

I have not found something about this in the [migration guide](https://image.nuxt.com/get-started/migration). (By the way, the link to this guide in the [release note](https://github.com/nuxt/image/releases/tag/v2.0.0) is broken.)

What am I missing?

image config in nuxt.config:

```
  image: {
    strapi: {
      baseURL: process.env.NUXT_PUBLIC_CMS_BASE_URL
    },
    cloudflare: {
      baseURL: process.env.NUXT_PUBLIC_ASSETS_BASE_URL
    }
  }
```

Nuxt version 3.16.2. previous nuxt/image version 1.11.0.

<img width="823" height="97" alt="Image" src="https://github.com/user-attachments/assets/80db47a9-ac76-4e23-b222-511a42552324" />

DamianGlowala:
Thanks for the heads up on broken links in the v2 release notes (all 4 of them are now correct).

PR #2141, opened 2026-02-26
fix: infer configured providers and their modifiers in `useImage`

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/image/issues/2036
resolves https://github.com/nuxt/image/issues/2107

### 📚 Description

Now, list of available providers is correctly autocompleted and there are no type errors when choosing a provider other than `ipx`. Also, the second argument (modifiers) now inferes available modifiers for a selected provider, so that they are now connected and type-safe.

```
const img = useImage()

img('<img_src>', { modifiers: <provider_specific> }, { provider: <provider> })
```

### DIFF
--- a/src/types/image.ts
+++ b/src/types/image.ts
@@ -82,15 +82,15 @@ export interface ImageSizes {
 }
 
 export interface Img {
-  <TProvider extends keyof ConfiguredImageProviders = keyof ConfiguredImageProviders>(
+  <TProvider extends keyof ConfiguredImageProviders>(
     source: string,
     modifiers?: ImageOptions<TProvider>['modifiers'],
     options?: ImageOptions<TProvider>
   ): ResolvedImage['url']
   options: CreateImageOptions
-  getImage: <TProvider extends keyof ConfiguredImageProviders = keyof ConfiguredImageProviders>(source: string, options?: ImageOptions<TProvider>) => ResolvedImage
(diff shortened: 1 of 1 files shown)


src/types/image.ts
```diff
@@ -19,17 +19,15 @@ export interface ResolvedImageModifiers extends ImageModifiers {
...
   densities?: string
   modifiers?: Partial<Omit<ImageModifiers, 'format' | 'quality' | 'background' | 'fit'>>
-    & ('modifiers' extends keyof ConfiguredImageProviders[Provider] ? ConfiguredImageProviders[Provider]['modifiers'] : Record<string, unknown>)
-  sizes?: string | Record<string, any>
-}
-
-export interface ImageSizesOptions extends ImageOptions {
-  sizes: Record<string, string | number> | string
+    & ('modifiers' extends keyof ConfiguredImageProviders[TProvider] ? ConfiguredImageProviders[TProvider]['modifiers'] : Record<string, unknown>)
+  sizes?: Sizes
```
danielroe, line 30: does this work in other contexts? e.g. if not passed to `getSizes`?

if it isn't respected then `ImageSizesOptions` should still be a separate type

DamianGlowala: from what I can see, `ImageSizesOptions` was used solely for the purpose of `getSizes`

PR #2143, opened 2026-02-26
fix!: require `src` prop in `<NuxtImg>` and `<NuxtPicture>`

DamianGlowala:
### 📚 Description

Marks `src` prop as required in `BaseImageProps` type (shared by both NuxtImg and NuxtPicture components).

### DIFF
--- a/src/runtime/components/NuxtImg.vue
+++ b/src/runtime/components/NuxtImg.vue
@@ -47,7 +47,7 @@ defineSlots<{ default(props: DefaultSlotProps): any }>()
 const $img = useImage()
 const { providerOptions, normalizedAttrs, imageModifiers } = useImageProps(props)
 
-const sizes = computed(() => $img.getSizes(props.src!, {
+const sizes = computed(() => $img.getSizes(props.src, {
   ...providerOptions.value,
   sizes: props.sizes,
   densities: props.densities,
@@ -86,7 +86,7 @@ const placeholder = computed(() => {
     ? placeholder
     : typeof placeholder === 'number' ? [placeholder] : []
 
-  return $img(props.src!, {
+  return $img(props.src, {
     ...imageModifiers.value,
     width,
     height,
@@ -98,7 +98,7 @@ const placeholder = computed(() => {
 const mainSrc = computed(() =>
   props.sizes
     ? sizes.value.src
-    : $img(props.src!, imageModifiers.value, providerOptions.value),
+    : $img(props.src, imageModifiers.value, providerOptions.value),
 )
 
 const src = computed(() => placeholder.value || mainSrc.value)
--- a/src/runtime/components/NuxtPicture.vue
+++ b/src/runtime/components/NuxtPicture.vue
@@ -78,7 +78,7 @@ const attrs = computed(() => {
   return attrs
 })
 
-const originalFormat = computed(() => props.src?.match(/^[^?#]+\.(\w+)(?:$|[?#])/)?.[1])
+const originalFormat = computed(() => props.src.match(/^[^?#]+\.(\w+)(?:$|[?#])/)?.[1])
 
 const legacyFormat = computed(() => {
   if (props.legacyFormat) {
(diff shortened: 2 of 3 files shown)


Bombastickj:
Hi,
I've just found this PR and unfortunately I think we should keep `src` optional and handle it within the component itself. Let's say for example I created my own custom image component for user profile pictures. When the user has no profile picture set, I want to display custom html (like an icon) and remove it when the image is set or after it has been loaded. This would not be pretty if the src is required.

Here is some example code (src optional):

```vue
<script setup lang="ts">
const {
  src = undefined,
  sizes = '32px',
  icon = 'tabler:photo',
} = defineProps<{
  src?: string | null;

  /** @default '32px' */
  sizes?: string;

  /** Icon name. @default 'tabler:photo' */
  icon?: string;
}>();
</script>

<template>
  <div>
    <NuxtImg
      v-slot="{ src: _src, isLoaded, imgAttrs }"
      :sizes
      :src="src || undefined"
      :custom="true"
    >
      <!-- Show the actual image when loaded -->
      <img v-if="isLoaded" v-bind="imgAttrs" :src="_src" />

      <Icon v-else :name="icon" />
    </NuxtImg>
  </div>
</template>
```

If src is required it would look something like this:

```vue
<script setup lang="ts">
const {
  src = undefined,
  sizes = '32px',
  icon = 'tabler:photo',
} = defineProps<{
  src?: string | null;

  /** @default '32px' */
  sizes?: string;

  /** Icon name. @default 'tabler:photo' */
  icon?: string;
}>();
</script>

<template>
  <div>
    <Icon v-if="!src" :name="icon" />

    <NuxtImg
      v-else
      v-slot="{ src: _src, isLoaded, imgAttrs }"
      :sizes
      :src="src || undefined"
      :custom="true"
    >
      <!-- Show the actual image when loaded -->
      <img v-if="isLoaded" v-bind="imgAttrs" :src="_src" />

      <Icon v-else :name="icon" />
    </NuxtImg>
  </div>
</template>
```

Or mabe like this:

```vue
<script setup lang="ts">
const {
  src = undefined,
  sizes = '32px',
  icon = 'tabler:photo',
} = defineProps<{
  src?: string | null;

  /** @default '32px' */
  sizes?: string;

  /** Icon name. @default 'tabler:photo' */
  icon?: string;
}>();

const loaded = ref(false);
</script>

<template>
  <div>
    <Icon v-if="!loaded" :name="icon" />

    <NuxtImg
      v-if="src"
      :sizes
      :src="src"
      @load="loaded = true"
    />
  </div>
</template>
```

Please tell me your thoughts on this, I hope my concern is understandable!

Bombastickj:
I have made a PR that handles this without a breaking change.

https://github.com/nuxt/image/pull/2218

DamianGlowala:
Hi @Bombastickj, thanks for sharing your concern and providing examples!

Having gone through them (and https://github.com/nuxt/image/pull/2218), I do believe the `src` prop should be required. Having it optional, in my opinion, encourages patterns I'd consider bad practice. `src` prop is essential for the `<NuxtImg>` to exist - without one e.g. there is nothing to load. I'd consider the second example the closest to what I'd implement myself (with some small tweaks, see below):
```
<div>
  <Icon v-if="!src" :name="icon" />

  <NuxtImg
    v-else
    v-slot="{ isLoaded, imgAttrs }"
    :sizes
    :src="src"
    :custom="true"
  >
    <!-- Show the actual image when loaded -->
    <img v-if="isLoaded" v-bind="imgAttrs" :src="src" />

    <Icon v-else :name="icon" />
  </NuxtImg>
</div>
```

I understand this might look like an unnecessary duplication of `<Icon>`, but that's still a perfectly manageable code; if your actual placeholder component is larger in size, I'd suggest either putting it in a standalone component or using https://vueuse.org/core/createReusableTemplate/#createreusabletemplate if you'd rather keep it in the same file.

Let me know your thoughts!

PR #2176, opened 2026-03-20
fix(nuxt-img): set `imagesrcset` preload when densities are set w/o sizes

vsergiu93:
### 🔗 Linked issue

no issue

### 📚 Description

We have a production issue where a preloaded image can be fetched twice.
When sizes is specified, the generated srcset uses width **descriptors (w)** instead of density **descriptors (x)**, so [hasMultipleDensities](https://github.com/nuxt/image/blob/v2.0.0/src/runtime/components/NuxtImg.vue#L107) becomes false. As a result, Nuxt Image generates a `<link rel="preload">` with **imagesizes** but without **imagesrcset**.

According to the [web.dev](https://web.dev/articles/preload-responsive-images) article on preloading responsive images, responsive image preloads should mirror the  attributes: src → href, srcset → imagesrcset, and sizes → imagesizes. The browser uses imagesrcset and imagesizes together for the same resource selection logic as srcset and sizes on ``

This change ensures that whenever the preload represents a responsive image, we also emit imagesrcset, not only when the srcset contains density descriptors. That keeps the preload aligned with the eventual `` request and avoids duplicate downloads.

I also added tests for this but not sure if they are added properly.

### DIFF
--- a/test/unit/bundle.test.ts
+++ b/test/unit/bundle.test.ts
@@ -21,7 +21,7 @@ describe.skipIf(process.env.ECOSYSTEM_CI || isWindows)('nuxt image bundle size',
       image: { provider: 'ipx' },
     })
 
-    expect(roundToKilobytes(withImage.totalBytes - withoutImage.totalBytes)).toMatchInlineSnapshot(`"12.2k"`)
+    expect(roundToKilobytes(withImage.totalBytes - withoutImage.totalBytes)).toMatchInlineSnapshot(`"12.6k"`)
   })
 })
 


test/unit/bundle.test.ts
```diff
@@ -21,7 +21,7 @@ describe.skipIf(process.env.ECOSYSTEM_CI || isWindows)('nuxt image bundle size',
       image: { provider: 'ipx' },
     })
 
-    expect(roundToKilobytes(withImage.totalBytes - withoutImage.totalBytes)).toMatchInlineSnapshot(`"12.6k"`)
+    expect(roundToKilobytes(withImage.totalBytes - withoutImage.totalBytes)).toMatchInlineSnapshot(`"12.2k"`)
```
vsergiu93: Not really sure why  this test failed for me locally, or why even it changed.

test/e2e/ssr.test.ts
```diff
@@ -55,6 +55,30 @@ describe('browser (ssr: true)', () => {
     })
   }
 
+  it('preloads image with correct attributes', async () => {
```
vsergiu93, line 58: Not sure if this is the best way to write this test, but I think it is important to have it.

test/unit/bundle.test.ts
vsergiu93: Just saw now that it fails now in github pipeline, I think I will just revert this change.

vsergiu93:
Hello @danielroe,

Please have a quick look over this one, curious if this PR makes sense for you and if it could be part of the next release. 

Thanks

PR #2218, opened 2026-04-23
fix: handle falsy `src` in `NuxtImg` and `NuxtPicture`

Bombastickj:
### 📚 Description
When `src: undefined` combined with `custom: true` & `sizes`, the preload link in `<head>` and the image load in `onMounted` can incorrectly resolve to `current-url/undefined`.

This causes `NuxtImg` to compute invalid values for the `sizes` computed, for example:

```
{
  "sizes": "32px",
  "src": undefined,
  "srcset": "undefined 32w, undefined 64w"
}
```

Which are then used within the component regardless of incorrect values.

Final edit:

   - Prevent `undefined` from being coerced into strings within `srcset` and `sizes`
   - Fix `getSizes` in `image.ts` to return empty values for falsy inputs
   - Remove unsafe non-null assertions in `NuxtImg` and `NuxtPicture`
   - Add parity test cases for both components to verify `src: undefined` handling


Bombastickj:
Just pushed a few more updates after a self-review, and I've added comprehensive test cases to ensure the new logic behaves as expected. Let me know if any further changes are needed once someone has the bandwidth to review!

PR #2296, opened 2026-07-30
ci: only run `release` workflows in nuxt org

luc122c:
### 📚 Description

The 'release' actions are running in forks. This PR ensures they only run in the Nuxt org. 

See https://github.com/hacknug/image/pull/1 for reference.


danielroe:
this is good

I've also fixed upstream in uppt

hacknug:
thanks luke!

